PR #77, opened 2016-03-18
Introduced debug flag

johanstokking:


htdvisser:
Debugging can be activated with the `TTN_DEBUG=true` environment variable. Why do we need a CLI flag for this?

PR #78, opened 2016-03-18
Move api in core folder

johanstokking:


johanstokking:
@htdvisser any idea what makes the test stuck and be killed after 10 minutes?

PR #94, opened 2016-03-25
Register user from ttnctl

johanstokking:
See #91


johanstokking:
We may want to force https here, or make it part of the `ttn-account-server` flag

johanstokking:
Obsolete since #95

PR #95, opened 2016-03-26
Login functionality for ttnctl

johanstokking:
Command `ttnctl user login [e-mail]` allows you to login. Tokens are persisted in `$HOME/.ttnctl/auths.json` and are available through `LoadAuth()` in the `utils` package

### DIFF
--- a/ttnctl/cmd/users.go
+++ b/ttnctl/cmd/users.go
@@ -4,29 +4,50 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 
+	"github.com/TheThingsNetwork/ttn/ttnctl/util"
 	"github.com/howeyc/gopass"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
 
+type token struct {
+	AccessToken      string `json:"access_token"`
+	Error            string `json:"error"`
+	ErrorDescription string `json:"error_description"`
+}
+
 // usersCmd represents the users command
 var usersCmd = &cobra.Command{
-	Use:   "users",
-	Short: "Manage users on the account server",
-	Long:  `ttnctl users allows you to manage users`,
+	Use:   "user",
+	Short: "Show the current user",
+	Long:  `ttnctl user shows the current logged on user`,
 	Run: func(cmd *cobra.Command, args []string) {
-		cmd.Help()
+		t, err := util.LoadAuth(viper.GetString("ttn-account-server"))
+		if err != nil {
+			ctx.WithError(err).Fatal("Failed to load authentication")
+		}
+
+		if t == nil {
+			ctx.Warn("No login found")
+			return
+		}
+
+		// TODO: Validate token
+
+		ctx.Infof("Logged on as %s", t.Email)
 	},
 }
 
 var usersCreateCmd = &cobra.Command{
 	Use:   "create [e-mail]",
 	Short: "Create a new user",
-	Long:  `ttnctl users create allows you to create a new user`,
+	Long:  `ttnctl user create allows you to create a new user`,
 	Run: func(cmd *cobra.Command, args []string) {
 		if len(args) < 1 {
 			ctx.Fatal("Insufficient arguments")
@@ -39,7 +60,7 @@ var usersCreateCmd = &cobra.Command{
 			ctx.Fatal("Invalid password")
 		}
 
-		uri := fmt.Sprintf("http://%s/register", viper.GetString("ttn-account-server"))
+		uri := fmt.Sprintf("%s/register", viper.GetString("ttn-account-server"))
 		values := url.Values{
 			"email":    {email},
 			"password": {string(password)},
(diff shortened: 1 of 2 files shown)


johanstokking:
Resolves #91

ttnctl/cmd/root.go
```diff
@@ -63,9 +63,8 @@ func init() {
 	RootCmd.PersistentFlags().String("app-eui", "0102030405060708", "The app EUI to use")
 	viper.BindPFlag("app-eui", RootCmd.PersistentFlags().Lookup("app-eui"))
 
-	RootCmd.PersistentFlags().String("app-token", "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJUVE4tSEFORExFUi0xIiwiaXNzIjoiVGhlVGhpbmdzVGhlTmV0d29yayIsInN1YiI6IjAxMDIwMzA0MDUwNjA3MDgifQ.zMHNXAVgQj672lwwDVmfYshpMvPwm6A8oNWJ7teGS2A", "The app Token to use")
-	viper.BindPFlag("app-token", RootCmd.PersistentFlags().Lookup("app-token"))
-
+	RootCmd.PersistentFlags().String("ttn-account-server", "https://account.thethings.network", "The address of the OAuth 2.0 server")
```
htdvisser: `account.thethingsnetwork.org`, right?

ttnctl/cmd/users.go
```diff
@@ -0,0 +1,145 @@
...
+		if err != nil {
+			ctx.WithError(err).Fatal("Failed to load authentication")
+		}
+
+		if t == nil {
+			ctx.Warn("No login found")
+			return
+		}
+
+		// TODO: Validate token
```
htdvisser: Maybe we should implement this before merging

ttnctl/cmd/users.go
```diff
@@ -0,0 +1,145 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) < 1 {
+			ctx.Fatal("Insufficient arguments")
+		}
+
+		email := args[0]
+		fmt.Print("Password: ")
+		password, err := gopass.GetPasswd()
+		if err != nil {
+			ctx.Fatal("Invalid password")
```
htdvisser: Errors from `gopass.GetPasswd()` usually mean an interrupt or read error. I don't think "Invalid password" is a clear description of the error here...

ttnctl/cmd/users.go
```diff
@@ -0,0 +1,145 @@
...
+		uri := fmt.Sprintf("%s/register", viper.GetString("ttn-account-server"))
+		values := url.Values{
+			"email":    {email},
+			"password": {string(password)},
+		}
+		res, err := http.PostForm(uri, values)
+		if err != nil {
+			ctx.WithError(err).Fatal("Registration failed")
+		}
+		defer res.Body.Close()
```
htdvisser: You're not even opening the Body..?

ttnctl/cmd/users.go
```diff
@@ -0,0 +1,145 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) < 1 {
+			ctx.Fatal("Insufficient arguments")
+		}
+
+		email := args[0]
+		fmt.Print("Password: ")
+		password, err := gopass.GetPasswd()
+		if err != nil {
+			ctx.Fatal("Invalid password")
```
htdvisser: Same as [here](https://github.com/TheThingsNetwork/ttn/pull/95/files#r57516677)

ttnctl/cmd/users.go
```diff
@@ -0,0 +1,145 @@
...
+			"grant_type": {"password"},
+			"username":   {email},
+			"password":   {string(password)},
+		}
+		req, err := http.NewRequest("POST", uri, strings.NewReader(values.Encode()))
+		if err != nil {
+			ctx.WithError(err).Fatal("Create request failed")
+		}
+		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+		req.SetBasicAuth("ttnctl", "")
```
htdvisser: Why?

ttnctl/cmd/users.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+type token struct {
+	AccessToken      string `json:"access_token"`
+	Error            string `json:"error"`
+	ErrorDescription string `json:"error_description"`
+}
+
+// usersCmd represents the users command
+var usersCmd = &cobra.Command{
```
htdvisser: `userCmd`

ttnctl/cmd/users.go
```diff
@@ -0,0 +1,145 @@
...
+				ctx.Fatalf("Request failed: %s", t.ErrorDescription)
+			} else {
+				ctx.Fatalf("Request failed: %s", resp.Status)
+			}
+		}
+
+		ctx.Infof("Logged in as %s", email)
+
+		if err := util.SaveAuth(server, email, t.AccessToken); err != nil {
+			ctx.WithError(err).Error("Failed to save login")
```
htdvisser: Could also be `Fatal`

ttnctl/util/auth.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"os/user"
+	"path"
+)
+
+const (
+	authsFileName = ".ttnctl/auths.json"
```
htdvisser: We might want to support setting the file location with a flag.

ttnctl/cmd/users.go
johanstokking: Tokens are valid, but the expiration is not checked. Will add that.

ttnctl/cmd/users.go
johanstokking: OAuth 2.0 requires basic authentication to identify the client with id as name and secret as password. Since ttnctl is a public app, a secret is not applicable.

ttnctl/util/auth.go
johanstokking: Why?

ttnctl/cmd/users.go
johanstokking: Expiration check and automatic refresh has been added

htdvisser:
I think it's a good idea to add unit tests for `ttnctl/cmd/user.go`

johanstokking:
What's that drone build failure?

htdvisser:
Looks like Drone had some trouble reaching Github. Re-ran it and everything's working now.

htdvisser:
Let's not use the default `$HOME/.ttnctl/auths.json` for testing.

PR #97, opened 2016-03-29
Manage applications from ttnctl

johanstokking:


htdvisser:
Will look into this after #95 is merged.

PR #99, opened 2016-03-29
Token validation in Broker

johanstokking:
Key is loaded from file `~/.ttn/oauth2-token.pub` but user can specify a different location using an environment variable or setting the `--oauth2-keyfile` flag. Because of the nature and size of this key, I think it should be a file and not putting the key as hex formatted string in an environment variable.

Note: the current PR breaks tests.

### DIFF
--- a/cmd/broker.go
+++ b/cmd/broker.go
@@ -5,6 +5,8 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
+	"path"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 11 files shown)


htdvisser:
@johanstokking: This PR contains #95 and #97 as well, maybe you should just close those and I'll focus on this one.

johanstokking:
This rebase wasn't necessary from a code dependency perspective, but this is all part of one big feature: support for authentication and authorization. Testing only works when all components talk tokens others can understand, and that goes much better when it's on one branch.

Note that the branches `feature/ttnctl-login`, `feature/ttnctl-applications` and `feature/broker-tokens` are incremental in that order. If you want to merge in chunks, do it that way. If you want to test integral, use this branch.

Op 30 mrt. 2016 om 00:20 heeft Hylke Visser <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


@johanstokking<https://github.com/johanstokking>: This PR contains #95<https://github.com/TheThingsNetwork/ttn/pull/95> and #97<https://github.com/TheThingsNetwork/ttn/pull/97> as well, maybe you should just close those and I'll focus on this one.

-

PR #108, opened 2016-03-30
[ttnctl] Add subscribe command

htdvisser:


ttnctl/cmd/subscribe.go
```diff
@@ -0,0 +1,134 @@
...
+		})
+
+		mqttClient := MQTT.NewClient(opts)
+
+		ctx.WithField("mqtt-broker", broker).Info("Connecting to The Things Network...")
+		if token := mqttClient.Connect(); token.Wait() && token.Error() != nil {
+			panic(token.Error())
+		}
+
+		topic := fmt.Sprintf("%X/devices/%s/up", appEUI, devEUI)
```
johanstokking: Use `ttn.core.api.topics` here

PR #114, opened 2016-03-31
Get public OAuth key from account server

### LINKED ISSUE #106: Request the public OAuth 2.0 key from the account server
This avoids provisioning the key on install of the Broker and allows for semi-automatic updates when the Broker restarts. We may want to add a management call to refresh the public key when the Broker runs, or automatic refresh on an interval. We need to be able to re-issue the key in a timely manner.

johanstokking:
Closes #106

### DIFF
--- a/cmd/broker.go
+++ b/cmd/broker.go
@@ -5,7 +5,6 @@ package cmd
 
 import (
 	"fmt"
-	"io/ioutil"
 	"path"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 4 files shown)


htdvisser:
The entire idea is for the Broker to be able to run even when the auth server is down. However, when the Broker starts, you call `Refresh()` which deletes the public key. A better approach would be:

```
on START:
if cachefile exists {
  read cachefile
  set key
}
do GET; if success {
  write cachefile
  set/update key
}
if key not set {
  fatal
}
```

Issue #110, opened 2016-03-30
Session keys after a join are incorrect


tftelkamp:
After a successful join, the nwkSkey and appSkey on the node differ from those in the backend.

htdvisser:
In the Handler they are calculated here: https://github.com/TheThingsNetwork/ttn/blob/develop/core/components/handler/handler.go#L464-L487

I don't immediately see a difference with the LoRaWAN spec. @KtorZ: do you see any mistake there?

htdvisser:
Same problem in `ttnctl join`: https://github.com/TheThingsNetwork/ttn/blob/develop/ttnctl/cmd/join.go#L145-L159

htdvisser:
Cause of this issue is probably the Endianness of the EUIs/SKeys

tftelkamp:
[LoRaWANandbyteorderingV3_923_2.xlsx](https://github.com/TheThingsNetwork/ttn/files/197128/LoRaWANandbyteorderingV3_923_2.xlsx)

htdvisser:
Fix is merged, will test this with @tftelkamp when he's available.

htdvisser:
Problem is solved.

PR #115, opened 2016-03-31
Fix for #110 - Calculate correct session keys in OTAA

htdvisser:


core/components/handler/handler.go
```diff
@@ -461,31 +461,28 @@ func (h component) consumeJoin(appEUI []byte, devEUI []byte, appKey [16]byte, da
...
-	binary.BigEndian.PutUint32(appNonce, rdn.Uint32())
+	var devAddr [4]byte
 	binary.BigEndian.PutUint32(devAddr[:], rdn.Uint32())
 	devAddr[0] = (h.Configuration.NetID[2] << 1) | (devAddr[0] & 1) // DevAddr 7 msb are NetID 7 lsb
 
-	buf := make([]byte, 16)
-	copy(buf[1:4], appNonce[:3])
-	copy(buf[4:7], h.Configuration.NetID[:])
-	copy(buf[7:9], packet.DevNonce)
+	// Generate appNonce - Note: this should be moved to the random package
```
johanstokking: Move indeed, but at least fill the 7 MSB with the NwkID (7 LSB of NetID, this should come from configuration)

htdvisser: See https://github.com/TheThingsNetwork/ttn/pull/115/files/bec6f174f938b742cd5fa1604b5e1e4bfa037195#diff-787ac1982e332ea58fbc8fc68bf4e182L470

Issue #104, opened 2016-03-30
Assign random security keys for devices


johanstokking:
When the user does not specify a security key when registering a device (either ABP or OTAA), generate random keys.

Keys should be generated by the Handler, as a convenience for the user.

PR #135, opened 2016-04-13
Default application key

### LINKED ISSUE #129: Support default AppKey for an application
The broker should accept activations by unknown devices using the default AppKey. When a device is registered, only that AppKey should be used. This avoids the need to register each single device.

johanstokking:
Closes #129


htdvisser:
Updated wrt #104. Merging after tests pass.

PR #141, opened 2016-04-15
Ask user to retype password to prevent typing errors when creating user

kersing:
Update to ttnctl utility to catch typos while creating accounts.

### DIFF
--- a/ttnctl/cmd/user.go
+++ b/ttnctl/cmd/user.go
@@ -4,10 +4,10 @@
 package cmd
 
 import (
+	"bytes"
 	"fmt"
 	"net/http"
 	"net/url"
-	"bytes"
 
 	"github.com/TheThingsNetwork/ttn/ttnctl/util"
 	"github.com/howeyc/gopass"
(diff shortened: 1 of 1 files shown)


ttnctl/cmd/user.go
```diff
@@ -49,6 +50,14 @@ var userCreateCmd = &cobra.Command{
 		if err != nil {
 			ctx.Fatal(err.Error())
 		}
+		fmt.Print("Retype password: ")
+		password2, err := gopass.GetPasswd()
```
johanstokking: Call it `confirmPassword`

ttnctl/cmd/user.go
```diff
@@ -49,6 +50,14 @@ var userCreateCmd = &cobra.Command{
...
+		if err != nil {
+			ctx.Fatal(err.Error())
+		}
+		if !reflect.DeepEqual(password, password2) {
```
johanstokking: Please use `bytes.Equal()` here? [docs](https://golang.org/pkg/bytes/#Equal)

johanstokking:
Thanks!

PR #142, opened 2016-04-18
Feature/core types

romeovs:
Add types for `AppEUI`, `DevEUI`, `DevAddr`, `AppKey`, `AppSKey` and `NwkSKey` to core.
Also adds JSON marshalling and un-marshalling for these types.


johanstokking:
Please replace `ttnctl.util` parse code by this, fix references and add simple test

htdvisser:
Please use `TextMarshaler` and `TextUnmarshaler` instead of JSON. 

```
type TextMarshaler interface {
        MarshalText() (text []byte, err error)
}
type TextUnmarshaler interface {
        UnmarshalText(text []byte) error
}
```

core/types.go
```diff
@@ -0,0 +1,166 @@
...
+	return parseHEX(input, 8)
+}
+
+// ParseEUI parses a 64-bit hex-encoded string
+func ParseEUI(input string) ([]byte, error) {
+	return parseHEX(input, 16)
+}
+
+// ParseKey parses a 128-bit hex-encoded string
+func ParseKey(input string) ([]byte, error) {
```
johanstokking: Should we be forward-compatible and introduce `ParseAppEUI()`, `ParseDevEUI()`, etc (same for `Key`s)? They might return `[]byte` so it doesn't break everything right away, but at least we have the naming right. Next step would be to change the `[]byte` to the concerning type alias.

romeovs: That would make integrating these type into the rest of the codebase a lot safer, and indeed be more future proof.

Maybe we should move the types to a separate module (such as `types`) so we have a bit more space to do this?

johanstokking: You might want to split type definitions in `types.go` and keep parsing and marshalling in other files, but it's all `core` package.

core/types.go
```diff
@@ -0,0 +1,190 @@
...
+	devAddr, err := hex.DecodeString(input)
+	if err != nil {
+		return nil, fmt.Errorf("Could not decode input: %s", input)
+	}
+
+	return devAddr, nil
+}
+
+// ParseAddr parses a 32-bit hex-encoded string
+func ParseAddr(input string) ([]byte, error) {
```
htdvisser: `ParseDevAddr` - We only use this for DevAddrs

core/types.go
```diff
@@ -0,0 +1,190 @@
...
+// ParseKey parses a 128-bit hex-encoded string
+func ParseKey(input string) ([]byte, error) {
+	return parseHEX(input, 32)
+}
+
+type AppEUI []byte
+type DevEUI []byte
+type DevAddr []byte
+
+type AppKey []byte
```
htdvisser: These keys can also be `lorawan.AES128Key`

core/types.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// ParseKey parses a 128-bit hex-encoded string
+func ParseKey(input string) ([]byte, error) {
+	return parseHEX(input, 32)
+}
+
+type AppEUI []byte
+type DevEUI []byte
+type DevAddr []byte
```
htdvisser: Can be `lorawan.DevAddr` 
See https://github.com/brocaar/lorawan/blob/master/fhdr.go#L11

core/types.go
```diff
@@ -0,0 +1,190 @@
...
+func ParseEUI(input string) ([]byte, error) {
+	return parseHEX(input, 16)
+}
+
+// ParseKey parses a 128-bit hex-encoded string
+func ParseKey(input string) ([]byte, error) {
+	return parseHEX(input, 32)
+}
+
+type AppEUI []byte
```
htdvisser: Can be `lorawan.EUI64`
See https://github.com/brocaar/lorawan/blob/master/payload.go#L11

htdvisser:
I really don't see why we need this feature. The types defined here are the same as the ones defined in the `lorawan` library we use.

johanstokking:
The Things Network is not LoRaWAN

Op 19 apr. 2016 om 11:28 heeft Hylke Visser <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


I really don't see why we need this feature. The types defined here are the same as the ones defined in the lorawan library we use.

-

htdvisser:
Will continue with this in `refactor` branch.

PR #143, opened 2016-04-24
Added developer mode for ABP devices

### LINKED ISSUE #118: Development mode for ABP devices
When ABP devices restart, their `FCnt` is reset. This results in messages being rejected.
When a device is configured for development mode, the backend should skip this `FCnt` check. Only question is what to do with downlink `FCnt`

kersing:
Added developer mode for ABP registered devices. Fixes #118


core/protos/handler.proto
```diff
@@ -10,6 +10,7 @@ message DataUpHandlerReq {
     uint32      FCnt        = 10;
     uint32      FPort       = 11;
     uint32      MType       = 12;
+    bool	FCUPRst	    = 13;
```
johanstokking: Make it `FCntUpReset`

johanstokking:
Looks good, thanks. We will test it asap

core/protos/handler.proto
kersing: Done.

KtorZ:
Doesn't look good to me.
Lot of change to introduce a boolean in a function only for debug. The debug should not change the underlying logic. 
Plus, what does this mean exactly?

```Go
		if entryCnt < (uint32(upperSup) - 10) {
			return devCnt, true, nil
		}
```

Where does that `-10` come from ?

KtorZ:
Also .... 
This test has no damn sense now:
https://github.com/kersing/ttn/blob/devmode/core/components/broker/controller_test.go#L266-L283

If the test doesn't pass after you added your feature, you'd rather change your feature to make it okay rather than changing the test :/

kersing:
Lot of changes to allow ABP devices that do not/can not store the frame counter to use TTN.

The check is to detect the FCntUP has been reset and it's not a roll-over of the 16 bit value. The -10 could have been -1 or any value (not too large).

KtorZ:
Then the test is wrong and this should be implemented in another way.

kersing:
The test checks if a reset is detected. The feature added is to check for reset/restarted devices. If the device does not start sending in range the counter value might be > 0, so the rewritten check validates wholeCounter returns the right value.

KtorZ:
The test was testing whether a upcoming frame counter was sent with a gap less than MAX_GAP (according to LoRaWAN specs). 
That test is and should still be valid after the feature is implemented. I see no reason to change it by something else.

kersing:
@KtorZ Suggestions for improvements welcome. I'm trying to help out, just posting 'this is wrong' does not encourage contributions.
Regarding the test, if the device is not set to developer mode the broker code will behave as previously. The test can not reflect this because it does not know if the device is set to developer mode.

core/components/broker/mocks_test.go
```diff
@@ -176,10 +176,10 @@ func (m *MockNetworkController) upsertNonces(entry noncesEntry) error {
 }
 
 // wholeCnt implements the NetworkController interface
-func (m *MockNetworkController) wholeCounter(devCnt, entryCnt uint32) (uint32, error) {
+func (m *MockNetworkController) wholeCounter(devCnt, entryCnt uint32) (uint32, bool, error) {
 	m.InWholeCounter.DevCnt = devCnt
 	m.InWholeCounter.EntryCnt = entryCnt
-	return m.OutWholeCounter.FCnt, m.Failures["wholeCounter"]
+	return m.OutWholeCounter.FCnt, false, m.Failures["wholeCounter"]
```
KtorZ: Not ok with this.
The pattern is straightforward, each output of mock structures are stored in the structure and are not hardcoded value. Therefore, we can change them on purpose for tests.

core/components/handler/handler.go
```diff
@@ -633,6 +634,9 @@ func (h component) consumeDown(appEUI []byte, devEUI []byte, dataRate string, bu
 			stats.MarkMeter("handler.downlink.pull")
 			downType := lorawan.UnconfirmedDataDown
 			ack := (upType == lorawan.ConfirmedDataUp)
+			if bundle.Packet.(*core.DataUpHandlerReq).FCntUpReset {
+				bundle.Entry.FCntDown = 0
```
KtorZ: This is overwritten 5 lines below.

core/components/broker/broker.go
```diff
@@ -270,6 +279,12 @@ func (b component) HandleData(bctx context.Context, req *core.DataBrokerReq) (*c
 		}
 
 		if ok {
+			if counterReset {
+				if !entry.DevMode {
+					continue
```
KtorZ: Doing this prevent the `mEntry` to be assigned + all the logging and analytic that happen below. 
This implies that in case of reset, no MIC check will ever succeed and the packet will just be ignored.

KtorZ:
By the way, it's not really possible to know whether the device has been reset. I mean, there's no way to know from the network point of view whether it's an error from the device or an actual reset. 
However, in order to implement that feature, the best way to go is seemingly by handling properly the error https://github.com/TheThingsNetwork/ttn/blob/develop/core/components/broker/broker.go#L255

Something like this for instance:

```Go
		fcnt32, err := b.NetworkController.wholeCounter(fcnt16, entry.FCntUp)
		if err != nil {
                        if !entry.DevMode {
                                continue
                        }
			// In dev mode, we assume the device is right in any case
                        fcnt32 = fcnt16
		}
```

core/components/handler/handler.go
kersing: After it has been used in h.buildDownlink to set downlink.Payload.MACPayload.FHDR.FCnt.

core/components/broker/mocks_test.go
kersing: Sorry about that, I did not know how to handle this properly. Didn't recognize the pattern and how to solve this the right way.

core/components/broker/broker.go
kersing: Eeehm, don't you think I tested this code with a life node and gateway? If a reset is 'detected' the counter will be set to the value as provided by the frame, the MIC check will succeed and the value of ok will be true. So if the counter has been reset and the device is not set to development mode the value gets rejected as it might be a failure of the device or a replay attack.

kersing:
@KtorZ First of all thank you for the feedback, this allows me to learn how TTN expects things to be done.

I agree there is no real way to check for a reset or a malfunction. However, the device has to be set development mode explicitly to allow this 'malfunction' so the 'owner' has to take deliberate steps. Also by checking the received counter is smaller then the expected value some malfunctions will be detected.

You solution is more elegant, however while working on this Johan suggested to check if the counter value received is smaller then expected so that's what I implemented. Your solution basically accepts any value. (Which might not be a bad thing) Your solution needs enhancement to reset the downlink counter as well, stated as requirement in issue 118.

KtorZ:
Hey,

Be sure that we're happy for having you contributing. However, we have to make sure that any new pieces of code submitted won't break things or doesn't break them. We don't expect you to right correctly right away, Should you not understand a choice that has been made, or a way to handle some logic, please ask, we'll be glad to guide you.

Thus, to sum up, I have problems merging this PR because:

- It overrides existing tests that *shouldn't be* overridden. 
- It does not add proper tests for the new feature introduced
- The `mock_controller` doesn't allow a tester to mock the new reply of `wholecounter` (even though I strongly think that the signature and the logic of this function shouldn't change)
- The implemented solution doesn't work properly and only handle cases where the counter is reset between the first 65535 packets. Should you try to reset after that, then you get unexpected behavior and no way to detect an error in non-dev mode.

I am pretty confident with the fact that you tested it with real nodes, yet we still need unit tests in the business logic. Also, you missed some edge-cases and this is precisely why tests stand for. Issues won't pop out from normal and expected behavior but rather from particular edge cases. 

Eventually, I didn't provide any solution for the problem, but merely a hint about what could be done in the broker, as a start, to tackle this problem. 

I hope this helps.

kersing:
@KtorZ 
To tackle the most fundamental issues first:
> The implemented solution doesn't work properly and only handles cases where the counter is reset within the first 65535 packets. Should you try to reset after that, then you get unexpected behavior and no way to detect an error in non-dev mode.

The limit of the first 65525 packets is a conscience design decision that can easily be changed. In my consideration for a _development_ node testing with that amount of packets should be sufficient. However, changing it to allow all frame counters less than the expected one is certainly possible by providing the mode of the device to wholeCounter and updating that function to something like:

```
// wholeCounter implements the broker.NetworkController interface
func (s *controller) wholeCounter(devCnt uint32, entryCnt uint32, bool devMode) (uint32, bool, error) {
        upperSup := int(math.Pow(2, 16))
        diff := int(devCnt) - (int(entryCnt) % upperSup)
        var offset int
        if diff >= 0 {
                offset = diff
        } else {
                if devMode {
                        return devCnt, true, nil
                }
                offset = upperSup + diff
        }
        if offset > upperSup/4 {
                return 0, false, errors.New(errors.Structural, "Gap too big, counter is errored")
        }
        return entryCnt + uint32(offset), false, nil
}
```

Would this solution address you issues and solve the edge-cases?

In non-dev mode the broker tries ValidateMIC with the new counter. That will succeed, however counterReset is set and devMode is not resulting in a continue at line 284 where previously it would continue at line 259.

I'll gladly update the code and work on the tests if we can agree on the basic solution.

johanstokking:
Thanks @kersing for contributing and @KtorZ for reviewing.

First, we're refactoring the components to get the separation of concerns right. In summary, it's moving network related functionality from Handler to Router, making the Broker performing deduplication and making the Network Server a first class component that remains node state. This means that most of this code will be moved or rewritten anyway, so let's not waste time on [gold plating](https://en.wikipedia.org/wiki/Gold_plating_(software_engineering)).

Second, I think that when the user enables development mode for a device, we can skip all frame counting checks and don't care for roll-over. When the FCntUp is less than the last known FCntUp, it resets both FCntUp and FCntDown. This has to happen after the message integrity check and is in my opinion not related. Currently, this is to tightly knitted.

Third, I think it's better to introduce a generic flags field for devices. We can use the bits to set features per device. This feature would be "allow frame counter resets" for ABP devices. The development mode is simply a bitmask setting this feature, but probably also future features that are simply defining bits. I think this saves us from going through all these files again when adding another boolean, plus that this specific feature may actually be useful in production for ABP devices that lose power but can't save state. This is common in unreliable power sources like a solar panel.

kersing:
Tried to update this branch to latest on develop. Merged version also available at https://github.com/kersing/ttn/tree/develop in case the update failed.

kersing:
For a version where the bool has been replaced with flags use [feature/bitflag](https://github.com/kersing/ttn/tree/feature/bitflag) in stead of this pull request.
That branch compiles but has _not_ been tested (yet).

johanstokking:
Thanks, I like the `bitflag` one better @kersing. Can you create a PR for that one? We can close this one then.

kersing:
@johanstokking  Created pull request 149 for bitflag version. Closing this one.

PR #150, opened 2016-04-29
Add DevEUI to MQTT data as seperate item

### LINKED ISSUE #145: Include DevEUI in MQTT uplink payload
If I subscribe to uplink messages for an application, I get the messages from the backend, containing a lot of information, like payload, port, counter and a lot of metadata, but the EUI of the originating device is missing. This can make it difficult to interpet (decode) messages, because the keys can be coupled to the device (like in the OTAA case).
If I subscribe to  +/devices/D0509916D0D10002/up I know to which device I'm connected, but if I subscribe to +/devices/+/up the originating devices cannot be deduced from the messages.

I would propose to put one of the device identifiers in the message (probably the DEVEUI)

kersing:
Implement feature request "Include DevEUI in MQTT uplink payload"

Resolves #145

### DIFF
--- a/core/adapters/mqtt/adapter.go
+++ b/core/adapters/mqtt/adapter.go
@@ -4,6 +4,8 @@
 package mqtt
 
 import (
+	"encoding/hex"
+	"strings"
 	"time"
 
 	"github.com/TheThingsNetwork/ttn/core"
@@ -39,6 +41,7 @@ func (a *defaultAdapter) HandleData(_ context.Context, req *core.DataAppReq, _ .
 		Metadata: core.ProtoMetaToAppMeta(req.Metadata...),
 		FPort:    uint8(req.FPort),
 		FCnt:     req.FCnt,
+		DevEUI:   strings.ToUpper(hex.EncodeToString(req.DevEUI)),
 	}
 
 	if a.ctx != nil {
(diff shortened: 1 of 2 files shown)


core/definitions.go
```diff
@@ -8,6 +8,7 @@ type DataUpAppReq struct {
 	Payload  []byte        `json:"payload"`
 	FPort    uint8         `json:"port,omitempty"`
 	FCnt     uint32        `json:"counter"`
+	DevEUI   string        `json:"deveui"`
```
johanstokking: Please use field name `dev_eui`

core/adapters/mqtt/adapter.go
```diff
@@ -39,6 +41,7 @@ func (a *defaultAdapter) HandleData(_ context.Context, req *core.DataAppReq, _ .
 		Metadata: core.ProtoMetaToAppMeta(req.Metadata...),
 		FPort:    uint8(req.FPort),
 		FCnt:     req.FCnt,
+		DevEUI:   strings.ToUpper(hex.EncodeToString(req.DevEUI)),
```
johanstokking: Please format with `fmt.Sprintf("%X", req.DevEUI)`

htdvisser:
Looks good to me.

PR #152, opened 2016-05-01
Check EUI for 'ttnctl applications use' command

### LINKED ISSUE #151: ttnctl should validate AppEUI when for applications use command
To prevent users from using a wrong AppEUI ttnctl should validate it for the 'applications use' command.

kersing:
Resolves #151.


johanstokking:
Looks good to me.

The `util.GetApplications()` should return a decent `AppEUI` so that you don't need to compare them as strings, but we're on that in another branch.

PR #160, opened 2016-05-13
Feature/collector

johanstokking:


core/adapters/fields/appStorage.go
```diff
@@ -0,0 +1,78 @@
...
+	_, err := client.Ping().Result()
+	if err != nil {
+		client.Close()
+		return nil, err
+	}
+	return &redisAppStorage{client}, nil
+}
+
+func makeKey(eui types.AppEUI) string {
+	return fmt.Sprintf(appKey, eui.String())
```
htdvisser: `%s` already formats as string

core/adapters/fields/appStorage_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	err := storage.SetFunctions(eui, functions)
+	a.So(err, ShouldBeNil)
+
+	fetchedFunctions, err := storage.GetFunctions(eui)
+	a.So(err, ShouldBeNil)
+	a.So(fetchedFunctions.Decoder, ShouldEqual, functions.Decoder)
+	a.So(fetchedFunctions.Converter, ShouldEqual, functions.Converter)
+	a.So(fetchedFunctions.Validator, ShouldEqual, functions.Validator)
+}
```
htdvisser: Please add a `TestGetFunction` and also check behaviour when functions are not set in Redis.

core/adapters/fields/functions_test.go
```diff
@@ -0,0 +1,97 @@
...
+	return data.humidity >= 0 && data.humidity <= 100;
+}`,
+	}
+
+	data, valid, err := functions.Process([]byte{40, 110})
+	a.So(err, ShouldBeNil)
+	a.So(valid, ShouldBeFalse)
+	a.So(data["temperature"], ShouldEqual, 20)
+	a.So(data["humidity"], ShouldEqual, 110)
+}
```
htdvisser: Please test behaviour when functions are not valid.

core/collection/appCollector.go
```diff
@@ -0,0 +1,74 @@
...
+// AppCollector represents a collector for application data
+type AppCollector interface {
+	Start() error
+	Stop()
+}
+
+type appCollector struct {
+	ctx     log.Interface
+	eui     types.AppEUI
+	broker  string
```
htdvisser: I think it's better to rename this to something like `mqttBroker`

core/collection/appCollector.go
```diff
@@ -0,0 +1,74 @@
...
+func (c *appCollector) handleUplink(client mqtt.Client, appEUI types.AppEUI, devEUI types.DevEUI, req core.DataUpAppReq) {
+	ctx := c.ctx.WithField("DevEUI", devEUI)
+
+	t, err := time.Parse(time.RFC3339, req.Metadata[0].ServerTime)
+	if err != nil {
+		ctx.WithError(err).Warnf("Invalid time: %v", req.Metadata[0].ServerTime)
+		return
+	}
+
+	err = c.storage.Save(appEUI, devEUI, t, req.Fields)
```
htdvisser: Maybe check if `req.Fields` is not `nil`

core/components/collector/appStorage.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"fmt"
+
+	"gopkg.in/redis.v3"
+
+	"github.com/TheThingsNetwork/ttn/core/types"
+)
+
+const (
```
htdvisser: I think `appKey` and `appsKey` can be very confusing

core/components/collector/collector.go
```diff
@@ -0,0 +1,127 @@
...
+	ac := collection.NewMqttAppCollector(ctx, c.broker, eui, key, c.dataStorage)
+	ctx.Info("Starting app collector")
+	if err := ac.Start(); err != nil {
+		return err
+	}
+	c.apps[eui] = ac
+	return nil
+}
+
+func (c *collector) stopApp(eui types.AppEUI) error {
```
htdvisser: `stopApp` is not tested.

core/components/collector/influxdb/influxdb.go
```diff
@@ -0,0 +1,54 @@
+// Copyright © 2016 The Things Network
```
htdvisser: Test coverage killer 😢

core/components/handler/handlerManager.go
```diff
@@ -35,6 +38,7 @@ func (h component) ListDevices(bctx context.Context, req *core.ListDevicesHandle
 	var abp []*core.HandlerABPDevice
 	var otaa []*core.HandlerOTAADevice
 	for _, dev := range entries {
+		// WTF?
```
htdvisser: Agreed.

ttnctl/cmd/root.go
```diff
@@ -18,6 +20,14 @@ var cfgFile string
 
 var ctx log.Interface
 
+func getHandlerManager() core.AuthHandlerClient {
```
htdvisser: Maybe do this in `util` package, similar to MQTT and Applications?

ttnctl/cmd/applicationsPayloadFunctions.go
```diff
@@ -0,0 +1,149 @@
...
+function by loading the specified files containing JavaScript code.
+`,
+	Run: func(cmd *cobra.Command, args []string) {
+		appEUI := util.GetAppEUI(ctx)
+
+		if len(args) == 0 {
+			cmd.Help()
+			return
+		}
+
```
htdvisser: This command is not very consistent. I would suggest to provide these options:

- `ttnctl applications pf set [decoder.js]` (for most use-cases only a decoder is enough)
- `ttnctl applications pf set [decoder.js] [converter.js]`
- `ttnctl applications pf set [decoder.js] [converter.js] [validator.js]` (all three at once)
- `ttnctl applications pf set --decoder decoder.js --converter converter.js --validator validator.js` (depending on which flags you add)

htdvisser:
This is quite a big PR, it would have been easier to split it up in two (conversion / collection).
In general I'm missing some unit tests for error cases (such as a not found in Redis).

htdvisser:
When running an end-to-end test, I get this panic:
```
ha.1   | panic: runtime error: invalid memory address or nil pointer dereference
ha.1   | 0x1 addr=0x8 pc=0xf37ec]
ha.1   | 0xc82000e200)
ha.1   | 481 +0x3e6
ha.1   | 0x2, 0x10, 0x0, 0x0, 0x0)
ha.1   |
ha.1   | github.com/TheThingsNetwork/ttn/core/adapters/fields.(*Functions).Process(0x0, 0xc820145990, 0x2, 0x10, 0x0, 0x1, 0x0, 0x0)
ha.1   | :93 +0x55
ha.1   | github.com/TheThingsNetwork/ttn/core/adapters/fields.(*adapter).HandleData(0xc820175a40, 0x11aa970, 0xc820078d98, 0xc820301810, 0x0, 0x0, 0x0, 0x445977c8, 0x0, 0x0)
ha.1   | 	$GOPATH/src/github.com/TheThingsNetwork/ttn/core/adapters/fields/adapter.go:56 +0xdd8
ha.1   |  +0x4c4
```

core/components/collector/appStorage.go
johanstokking: Yeah, and add `AppSKey` to that list. Renaming this one to AccessKey

ttnctl/cmd/applicationsPayloadFunctions.go
johanstokking: The decoder function is mandatory, both the converter and the validator function are optional so they are flags. Therefore, bullet 2-4 doesn't work.

johanstokking:
Sure it's a big one. The conversion was first part of the collection, then it was moved to the handler and that made it hard to split.

What do you mean with not found in Redis btw? These payload functions are all optional; if the decoder does not exist, it just doesn't perform them and everything works as if it used to work.

htdvisser:
> What do you mean with not found in Redis

appStorage.go:54-60:
```
func (s *redisAppStorage) GetFunctions(eui types.AppEUI) (*Functions, error) {
	m, err := s.client.HGetAllMap(makeKey(eui)).Result()
	if err == redis.Nil {
		return nil, nil
	} else if err != nil {
		return nil, err
	}
```

adapter.go:56:
```
	fields, valid, err := functions.Process(in.Payload)
```

On Redis not found, `functions` is `nil`, causing a panic. This could have been detected with a unit test.

PR #161, opened 2016-05-17
add IsEmpty to core types

romeovs:
This pull request adds the `IsEmpty` method on all core types. `IsEmpty` is a helper function that checks if the value under consideration is the default value for this type (for instance `000000000000000` for an `AppEUI`).

### DIFF
--- a/core/types/dev_addr_test.go
+++ b/core/types/dev_addr_test.go
@@ -57,4 +57,9 @@ func TestDevAddr(t *testing.T) {
 	err = uOut.Unmarshal(bin)
 	a.So(err, ShouldBeNil)
 	a.So(*uOut, ShouldEqual, addr)
+
+	// IsEmpty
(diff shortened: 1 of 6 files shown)


core/types/dev_addr_test.go
```diff
@@ -57,4 +57,9 @@ func TestDevAddr(t *testing.T) {
 	err = uOut.Unmarshal(bin)
 	a.So(err, ShouldBeNil)
 	a.So(*uOut, ShouldEqual, addr)
+
+	// IsEmpty
+	var empty DevAddr
+	a.So(empty.IsEmpty(), ShouldEqual, true)
+	a.So(addr.IsEmpty(), ShouldEqual, false)
```
johanstokking: Just for readability; `ShouldBeTrue`/`ShouldBeFalse`

PR #166, opened 2016-05-19
[mqtt] Retry Connect

htdvisser:


mqtt/client.go
```diff
@@ -114,12 +114,28 @@ func NewClient(ctx log.Interface, id, username, password string, brokers ...stri
...
-	if token := c.mqtt.Connect(); token.Wait() && token.Error() != nil {
-		return fmt.Errorf("Could not connect: %s", token.Error())
+	for retries := 0; retries < ConnectRetries; retries++ {
+		token := c.mqtt.Connect()
+		token.Wait()
+		err = token.Error()
+		if err == nil {
+			break
+		}
+		<-time.After(time.Duration(retries) * ConnectRetryDelay)
```
johanstokking: `ConnectRetryDelay`'s comment indicates that it is the time to wait between retries, but here you multiply it by the number of retries so far. Either update command or get rid of multiplication. Also, on the first retry, `retries` is 0 so there's no wait

Issue #155, opened 2016-05-09
EU RX2 parameters


tftelkamp:
LoRaWAN specification 1.0/1.0.1:

"The RX2 receive window uses a fixed frequency and data rate. The default parameters are 869.525 MHz / DR0 (SF12, 125 kHz)."

For scalability reasons TTN wil use SF9 as RX2 data rate (already implemented). End-nodes need to be (made) aware of that.

** OTAA **

The RX2 data rate configuration is communicated to the end-node in the join accept message (DLsettings). After the join, the node can be assumed to be aware of the RX2 SF9 rate. If the stack, e.g. LMiC, does not process the DLsettings correctly, then it needs to be configured manually on the end-node (after the join).

Before the join, the end-node still uses the default RX2 data rate (SF12). To be fully compliant with the specification, the TTN back-end needs to use SF12 in the RX2 slot during the join process. This alters the downlink logic from the current implementation.

join request SF7-SF11: join accept in RX1 (SF7-SF11)
join request SF12: join accept in RX2 (SF12)

** ABP **

ABP end-nodes need to be configured manually to use SF9 as RX2 data rate. 

The TTN backend could try to facilitate this by sending MAC commands to set the correct configuration, for example after a frame counter reset. TBD.

avbentem:
Or maybe, for join accept, SF12 can be used in RX1 as well?

(Indeed SF12 needs to be used in RX2, if RX1 was not used for some reason. But SF12 being mandatory for RX2 join accepts, might not necessarily mean that SF12 cannot be used in RX1 as well?)

tftelkamp:
Due to the frequency plan, RX1 has a duty cycle of 1% and RX2 of 10%. Better to schedule long SF12 transmissions in RX2. Also, if the RX1 duty cycle is (near) full already, the join accept will be in RX2/SF12 anyhow.

jverhoeven:
Can we label this as 'bug'. I wonder how people can actually use OTAA with TTN backend with this deviation from the spec. At least my libmDot does not allow me to change the SF for RX2 when it expects a JoinAccept message.

kersing:
If the consensus is to have join reply use the standard setting of SF12 in RX2, this should be implemented sooner rather than later. People are starting to build nodes for deployment, if this is changed after deployment those nodes will require firmware updates for the new setting. (Or ugly work arounds need to be added to firmware now)

htdvisser:
[...]

PR #176, opened 2016-06-05
Fix for join issues mentioned in "EU RX2 parameters #155", …

kersing:
Modified the code to make sure JoinAccepts in RX2 get transmitted using SF12 as per LoRaWAN standard. Did not change scoreComputer to schedule response packets for SF7-SF11 in RX1 as it does not make sense given Thomas's remark concerning duty cycles of those channels.

As this patch does not fix the mentioned LMiC issues nor the requirement to set RX2 to SF9 for ABP nodes.

### DIFF
--- a/core/components/handler/handler.go
+++ b/core/components/handler/handler.go
@@ -48,6 +48,7 @@ type component struct {
 		NetID       [3]byte
 		RX1DROffset uint8
 		RX2DataRate string
+		RX2JoinRate string
 		RX2Freq     float32
 		RXDelay     uint8
 		PowerRX1    uint32
@@ -111,6 +112,7 @@ func New(c Components, o Options) Interface {
 	h.Configuration.NetID = [3]byte{14, 14, 14}
 	h.Configuration.RX1DROffset = 0
 	h.Configuration.RX2DataRate = "SF9BW125"
+	h.Configuration.RX2JoinRate = "SF12BW125"
 	h.Configuration.RX2Freq = 869.525
 	h.Configuration.RXDelay = 1
 	h.Configuration.JoinDelay = 5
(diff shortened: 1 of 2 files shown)


jverhoeven:
Thanks for making the fix. Just tested it and my OTA join requests work on my mDot based node.

htdvisser:
Looks good, @kersing!

It would be even more awesome if you could also change this in the `feature/us-frequencies` branch (which runs on _staging_). Would you have time to do this as well?

tftelkamp:
I tested it, and it works fine.

However, the logic is broken now: we answer (SF9-SF11) join requests with a lower data rate, hence more airtime.

It would be much better if SF7-SF11 join accepts get responses in RX1, if duty-cycle allows.

For RX2 downlink message (at SF9), the logic is unchanged.

jverhoeven:
Now that you mention it, I always get an answer in RX2. What is the reason I never get any responses in RX1? I run my own (multi-tech) gateway with hardly any users, so I would say there should be sufficient bandwidth on RX1.

tftelkamp:
It mostly depends on the join request. An SF7 join will most likely get a response in RX1, unless the duty cycle limit has been reached.

kersing:
@jverhoeven In my tests with MultiTech gateway (with polyfwd) join requests using SF7 and SF8 get response in RX1.

kersing:
@tftelkamp Hmm, done some testing and using SF10 & SF11 with RX1 might not be a good idea. Got lots of denied messages because gateway could not transmit scheduled join-accept because of duty cycle limit.
Accept @ SF11 takes almost 750 ms which means a block for 75 secs. Using RX2 blocks that frequency for just 15 seconds.

BTW. shouldn't the dutycycle manager in the back end avoid using RX1 for the next 75 seconds??

Issue #127, opened 2016-04-05
Add port for mqtt downlink


tftelkamp:
Application should be able to specify port for downlink messages, either via the topic or json.

KtorZ:
What do you mean exactly? 
Application being able to connect to the mqtt server on the port they want ?

htdvisser:
Setting the LoRaWAN FPort

htdvisser:
This is relatively easy to implement, but requires a migration of the database in which packets are scheduled. As this database is not easy to migrate, it might be better to wait until we switch from our current binary datastore (Bolt) to Redis.

htdvisser:
This has been implemented in the `refactor` branch. See #199 for more info.

PR #184, opened 2016-06-10
Add port to MQTT json messages. Implements #127.

kersing:
Patch allows setting port on downlink messages.


core/components/handler/pktStorage.go
```diff
@@ -145,6 +147,7 @@ func (e pktEntry) MarshalBinary() ([]byte, error) {
 	rw := readwriter.New(nil)
 	rw.Write(e.AppEUI)
 	rw.Write(e.DevEUI)
+	rw.Write(uint16(e.FPort))
```
htdvisser: This line makes it difficult to deploy this feature to our _staging_ environment, as existing databases can still contain data in the old format. This means we have three options:

- Don't deploy this to staging
- Delete all old data and deploy to staging
- Write something that migrates old data

I vote for option 2, as it is the simplest solution. There aren't that many scheduled packets that would get lost

kersing: I vote for option 2 as well as we're currently not able to send any downlink data anyway... So deploying it now would not make a difference.

htdvisser: What do you mean with "currently not able to send any downlink data anyway"?
The current implementation just sets the downlink port to 1, which is working just fine.

kersing: see http://forum.thethingsnetwork.org/t/downlink-problem/2433 - Seems to be an issue with downlink at the moment.

htdvisser:
We will not merge this PR at this time because of the issues this will cause with the staging environment. Support for a downlink FPort has been implemented in the `refactor` branch (#199) and will be deployed in the production environment.

Nevertheless I would like to thank @kersing for his work.

PR #196, opened 2016-07-06
add helper to convert DevAddr to DevEUI

romeovs:
Adds a helper fn on the type `DevAddr` to create a `DevEUI` from it:

```go
addr := types.DevAddr{1,2,3,4}
eui := addr.ToEUI()
// is the same as
eui := types.DevEUI{0, 0, 0, 0, 1, 2, 3, 4}
```

### DIFF
--- a/core/types/dev_addr.go
+++ b/core/types/dev_addr.go
@@ -80,10 +80,11 @@ func (addr DevAddr) IsEmpty() bool {
 	return addr == empty
 }
 
-var prefix = []byte{0, 0, 0, 0}
+// ABPEUIPrefix is the prefix denoting a prefixed dev address
+var ABPEUIPrefix = []byte{0, 0, 0, 0}
 
 func (addr DevAddr) ToEUI() DevEUI {
 	var devEUI DevEUI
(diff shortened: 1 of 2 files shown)


htdvisser:
LGTM

core/types/dev_addr.go
```diff
@@ -79,3 +79,11 @@ var empty DevAddr
 func (addr DevAddr) IsEmpty() bool {
 	return addr == empty
 }
+
+var prefix = []byte{0, 0, 0, 0}
```
johanstokking: `prefix` is a bit too generic here

htdvisser: @romeovs: If you decide to change this, you can also just do:

```go
func (addr DevAddr) ToEUI() DevEUI {
	var devEUI DevEUI
	copy(devEUI[4:], addr.Bytes()))
	return devEUI
 }
```

johanstokking:
I don't want to start muggenzifting, but I think it's better to be explicit about that prefix instead of relying on a zero filled byte slice. There are more prefixes allowed without colliding with actual `DevEUI`s from MAC blocks

Issue #209, opened 2016-08-01
Get Gateway Status from Router


htdvisser:
The Router should expose the latest status received from a gateway through the [`GatewayStatus` RPC](https://github.com/TheThingsNetwork/ttn/blob/refactor/api/router/router.proto#L98) in the `RouterManager`.

Please add a command to `ttnctl`that retrieves the status of the gateway from the Router and prints it to the command line.

b1101:
closed by #212

PR #211, opened 2016-08-02
Gateway status command implementation

b1101:
See #209


b1101:
did not check gps for 0,0 will add a commit in a sec

PR #225, opened 2016-08-31
Bugfix/encoder types

romeovs:
Fixes a bug where the encoding of payloads would not work as expected because of mismatched integer types. For instance, the encoder:
```js
function Encoder(fields) {
  return fields.foo;
}
```

With `fields = [1, 2, 3] ` returns the error `Encoder should return an Array of numbers`. This is not correct because you'd expect this to work (all values in the array are integers between 0 an 255).

This PR fixes that by allowing more types (`byte`, `int`, `int32`, `uint32`, `int64`, `uint64`, `float32` and `float64`) to be returned. In case of the floating point types, it returns an error when the value is not an integer value.

### DIFF
--- a/core/handler/convert_fields.go
+++ b/core/handler/convert_fields.go
@@ -234,7 +234,18 @@ func (f *DownlinkFunctions) Encode(payload map[string]interface{}) ([]byte, erro
 			n = int64(t)
 		case uint64:
 			n = int64(t)
+		case float32:
+			n = int64(t)
+			if float32(n) != t {
+				return nil, errors.New("Encoder should return an Array of integer numbers")
+			}
+		case float64:
+			n = int64(t)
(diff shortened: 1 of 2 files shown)


core/handler/convert_fields.go
```diff
@@ -207,17 +208,54 @@ func (f *DownlinkFunctions) Encode(payload map[string]interface{}) ([]byte, erro
...
+	s := reflect.ValueOf(v)
+	l := s.Len()
+
+	res := make([]byte, l)
+
+	var n int64
+	for i := 0; i < l; i++ {
+		el := s.Index(i).Interface()
+
+		// type switch does not have falltrought so we need
```
johanstokking: fallthrough

johanstokking:
Looks good

PR #226, opened 2016-09-02
Feature/dry run

romeovs:
Adds the ability to dry run payload functions on the handler without having to save them or send actual messages. Very handy for debugging.

I called it dry run instead of *payload functions test* or something like that because we might want to
pull in more message transformations in the long run.

Please discuss if you like the idea/API or not, I will implement a `ttnctl` command for it then.


johanstokking:
Looks good to me

htdvisser:
Looks good to me 👍

PR #235, opened 2016-09-08
Feature/check app rights

romeovs:
- Generalizes `Account` to be able to authenticate with access keys (or be used without authentication).
- Implements `AppRights` with returns the rights the `Account` has to a certain app.


core/account/util/http.go
```diff
@@ -44,21 +46,20 @@ func checkRedirect(req *http.Request, via []*http.Request) error {
...
 
-	req.Header.Set("Authorization", fmt.Sprintf("bearer %s", accessToken))
 	req.Header.Set("Accept", "application/json")
 	req.Header.Set("Content-Type", "application/json")
 
 	return req, nil
 }
 
-func performRequest(server, accessToken, method, URI string, body, res interface{}, redirects int) (err error) {
+func performRequest(server string, s auth.Strategy, method, URI string, body, res interface{}, redirects int) (err error) {
```
johanstokking: Rename to `strategy`

romeovs: done

PR #237, opened 2016-09-11
Encoder/example

AntoineRondelet:
I add an example of encoder function in `ttnctl/cmd/applications_pf_set.go`

### DIFF
--- a/ttnctl/cmd/applications_pf_set.go
+++ b/ttnctl/cmd/applications_pf_set.go
@@ -90,13 +90,14 @@ The functions are read from the supplied file or from STDIN.`,
 ########## Write your Validator here and end with Ctrl+D (EOF):`)
 				app.Validator = readFunction()
 			case "encoder":
-				// val is the JSON payload we want to encode.
-				fmt.Println(`function (val) {
-	var str = JSON.stringify(val)
-	return Array.prototype.slice.apply(str).map(function (x) {
-		// todo: return an array of numbers
-		return x.charCodeAt(0)
-	})
+				fmt.Println(`function encoder(jsonPayload) {
+	var str = JSON.stringify(jsonPayload);
+	strTab = Array.prototype.slice.apply(str);
+	var bytePayload = [];
+	for(i=0; i<strTab.length; i++) {
+		bytePayload.push(strTab[i].charCodeAt(0));
+	}
+	return bytePayload;
 }
 ########## Write your Encoder here and end with Ctrl+D (EOF):`)
 				app.Encoder = readFunction()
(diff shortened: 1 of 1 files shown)


ttnctl/cmd/applications_pf_set.go
```diff
@@ -90,11 +90,13 @@ The functions are read from the supplied file or from STDIN.`,
 ########## Write your Validator here and end with Ctrl+D (EOF):`)
 				app.Validator = readFunction()
 			case "encoder":
+				// val is the JSON payload we want to encode.
```
johanstokking: Remove

ttnctl/cmd/applications_pf_set.go
```diff
@@ -90,11 +90,13 @@ The functions are read from the supplied file or from STDIN.`,
...
 				fmt.Println(`function (val) {
-  // val is the output of the encoder function.
-
-  // todo: return an array of numbers
-  return return [96, 4, 3, 2, 1, 0, 1, 0, 1, 0, 0, 0, 0];
+	var str = JSON.stringify(val)
+	return Array.prototype.slice.apply(str).map(function (x) {
+		// todo: return an array of numbers
+		return x.charCodeAt(0)
+	})
```
johanstokking: Just remove one `return`. Most users are not going to understand this example

ttnctl/cmd/applications_pf_set.go
```diff
@@ -90,11 +90,14 @@ The functions are read from the supplied file or from STDIN.`,
...
 				app.Validator = readFunction()
 			case "encoder":
-				fmt.Println(`function (val) {
-  // val is the output of the encoder function.
-
-  // todo: return an array of numbers
-  return return [96, 4, 3, 2, 1, 0, 1, 0, 1, 0, 0, 0, 0];
+				fmt.Println(`function encoder(jsonPayload) {
+	var str = JSON.stringify(jsonPayload);
+	strTab = Array.prototype.slice.apply(str);
```
FokkeZB: As @johanstokking already said stringify the full payload is probably not a best practice, but it doesn't seem to actually work either, or at least a test with this code fails in the dashboard with:

```js
Error("Could not dry-run downlink with fields on Handler: rpc error: code = 2 desc = Encoder not valid: should return an Array of integer numbers")
```

Which is weird because it does return an array of numbers.

Also using this version:

```js
function Encoder(obj) {
	var str = obj.greeting; // 'Hi'
	var bytePayload = [];
	var strTab = Array.prototype.slice.apply(str);
	for (i = 0; i < strTab.length; i++) {
		bytePayload.push(strTab[i].charCodeAt(0));
	}
	return bytePayload;
}
```

Or the more simple:

```js
function Encoder(obj) {
	var str = obj.greeting; // 'Hi'
	return str.split('').map(function(char) {
		return char.charCodeAt();
	});
}
```

PR #240, opened 2016-09-12
Add extra types to pf Encode

### LINKED ISSUE #239: Payload Functions: Encoder test fails while function is correct
If I use this for Encoder:

```js
function Encoder(obj) {
	return "Hi".split('').map(function(char) {
		return char.charCodeAt();
	});
}
```

Test fails with:

```
Encoder not valid: should return an Array of integer numbers
```

But run this in the browser or Node.js and you do get an Array of integer numbers:
```js
[ 72, 105 ]
```

romeovs:
Fixes #239

### DIFF
--- a/core/handler/convert_fields.go
+++ b/core/handler/convert_fields.go
@@ -221,12 +221,20 @@ func (f *DownlinkFunctions) Encode(payload map[string]interface{}) ([]byte, erro
 	for i := 0; i < l; i++ {
 		el := s.Index(i).Interface()
 
+		fmt.Printf("VAL %v TYPE %T\n", el, el)
+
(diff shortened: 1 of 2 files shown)


core/handler/convert_fields.go
```diff
@@ -221,13 +221,21 @@ func (f *DownlinkFunctions) Encode(payload map[string]interface{}) ([]byte, erro
 	for i := 0; i < l; i++ {
 		el := s.Index(i).Interface()
 
+		fmt.Printf("VAL %v TYPE %T\n", el, el)
```
johanstokking: Remove

PR #246, opened 2016-09-16
Feature/account server gateway

romeovs:
Adds client to `core/account` to manage gateways on the account server.

### DIFF
--- a/core/account/gateways.go
+++ b/core/account/gateways.go
@@ -1,0 +1,94 @@
+// Copyright © 2016 The Things Network
+// Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+
+package account
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/TheThingsNetwork/ttn/core/types"
+)
+
+// ListApplications list all applications
+func (a *Account) ListGateways() (gateways []Gateway, err error) {
+	err = a.get("/gateways", &gateways)
+	return gateways, err
+}
+
+// FindGateway returns the information about a specific gateay
+func (a *Account) FindGateway(gatewayID string) (gateway Gateway, err error) {
+	err = a.get(fmt.Sprintf("/gateways/%s", gatewayID), &gateway)
+	return gateway, err
+}
+
+// NewGateway is used as a paramater to CreateGateway to allow for optional
+// arguments
+type NewGateway struct {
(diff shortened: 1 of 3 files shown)


core/account/types.go
```diff
@@ -59,3 +59,22 @@ type Component struct {
...
 }
+
+// Gateway represents a gateway on the account server
+type Gateway struct {
+	ID               string           `json:"id" valid:"required"`
+	EUI              types.GatewayEUI `json:"eui" valid:"required"`
+	Activated        bool             `json:"activated"`
+	FrequencyPlan    string           `json:"frequency_plan"`
+	FrequencyPlanURL string           `json:"frequency_plan_url"`
+	PublicLocation   bool             `json:"location_public"`
```
htdvisser: Be consistent. `StatusPublic` vs `PublicLocation`.

core/account/types.go
```diff
@@ -59,3 +59,22 @@ type Component struct {
...
+	StatusPublic     bool             `json:"status_public"`
+	Location         *Location        `json:"location"`
+	Collaborators    []Collaborator   `json:"collaborator"`
+	Key              string           `json:"key"`
+}
+
+type Location struct {
+	Lng float64 `json:"lng"`
+	Lat float64 `json:"lat"`
+}
```
htdvisser: Please don't shorten `Lat` and `Lng` in the struct.

core/account/gateways.go
```diff
@@ -0,0 +1,112 @@
...
+
+// ChangeFrequencyPlan changes the requency plan of a gateway
+func (a *Account) ChangeFrequencyPlan(gatewayID, plan string) (Gateway, error) {
+	return a.EditGateway(gatewayID, GatewayEdits{
+		FrequencyPlan: plan,
+	})
+}
+
+// ChangeLocation changes teh location of the gateway
+func (a *Account) ChangeLocation(gatewayID string, location Location) (Gateway, error) {
```
htdvisser: Maybe just make this `ChangeLocation(gatewayID string, latitude, longitude float64)` ?

romeovs: The problem there is that its is easy to forget which one comes first, so the API becomes less clear.

htdvisser: Don't people use godoc/gocode/godef anymore?

PR #261, opened 2016-09-19
core/router/gateways: NOC integration(WIP)

rvolosatovs:
NOC integration in TTN

Progress:
- [x] `core/router/gateway` integration
- [ ] testing
- [ ] initialize NOC connection
- [ ] use `sync.RWMutex` instead of `sync.Mutex` in `core/router/gateway` for a possible performance benefit(?)
- [ ] token validation


core/router/server.go
```diff
@@ -21,39 +21,65 @@ type routerRPC struct {
...
-	return
+	return token[0], nil
 }
 
 // GatewayStatus implements RouterServer interface (github.com/TheThingsNetwork/ttn/api/router)
 func (r *routerRPC) GatewayStatus(stream pb.Router_GatewayStatusServer) error {
-	gatewayID, err := getGatewayFromMetadata(stream.Context())
+	md, err := metadataFromContext(stream.Context())
+
+	id, err := gatewayFromMetadata(md)
```
htdvisser: We usually write `gtwID` or `gatewayID`

core/router/server.go
```diff
@@ -65,16 +91,27 @@ func (r *routerRPC) GatewayStatus(stream pb.Router_GatewayStatusServer) error {
...
+		go r.router.HandleGatewayStatus(id, status)
 	}
 }
 
 // Uplink implements RouterServer interface (github.com/TheThingsNetwork/ttn/api/router)
 func (r *routerRPC) Uplink(stream pb.Router_UplinkServer) error {
-	gatewayID, err := getGatewayFromMetadata(stream.Context())
+	md, err := metadataFromContext(stream.Context())
+
+	id, err := gatewayFromMetadata(md)
```
htdvisser: We usually write `gtwID` or `gatewayID`

core/router/server.go
```diff
@@ -86,21 +123,23 @@ func (r *routerRPC) Uplink(stream pb.Router_UplinkServer) error {
...
 		}
-		go r.router.HandleUplink(gatewayID, uplink)
+		go r.router.HandleUplink(id, uplink)
 	}
 }
 
 // Subscribe implements RouterServer interface (github.com/TheThingsNetwork/ttn/api/router)
 func (r *routerRPC) Subscribe(req *pb.SubscribeRequest, stream pb.Router_SubscribeServer) error {
-	gatewayID, err := getGatewayFromMetadata(stream.Context())
+	id, err := gatewayFromContext(stream.Context())
```
htdvisser: We usually write `gtwID` or `gatewayID`

core/router/server.go
```diff
@@ -118,10 +157,11 @@ func (r *routerRPC) Subscribe(req *pb.SubscribeRequest, stream pb.Router_Subscri
 
 // Activate implements RouterServer interface (github.com/TheThingsNetwork/ttn/api/router)
 func (r *routerRPC) Activate(ctx context.Context, req *pb.DeviceActivationRequest) (*pb.DeviceActivationResponse, error) {
-	gatewayID, err := getGatewayFromMetadata(ctx)
+	id, err := gatewayFromContext(stream.Context())
```
htdvisser: We usually write `gtwID` or `gatewayID`

core/router/uplink.go
```diff
@@ -15,9 +15,8 @@ import (
 	"github.com/brocaar/lorawan"
 )
 
-func (r *router) HandleUplink(gatewayID string, uplink *pb.UplinkMessage) error {
+func (r *router) HandleUplink(gatewayID string, uplink *pb.UplinkMessage) (err error) {
 	ctx := r.Ctx.WithField("GatewayID", gatewayID)
-	var err error
```
htdvisser: Removing error handling here, see earlier comment

core/router/gateway/gateway.go
```diff
@@ -26,4 +31,71 @@ type Gateway struct {
...
+		}()
+	}
+
+	g.updateTimestamp()
+	g.Schedule.Sync(uplink.GatewayMetadata.Timestamp)
+	return g.Utilization.AddRx(uplink)
+}
+
+func (g *Gateway) HandleDownlink(identifier string, downlink *pb_router.DownlinkMessage) (err error) {
+	if g.monitor != nil {
```
htdvisser: We probably want to do this after we get the result, because at this point the func might still fail

core/router/gateway/gateway.go
```diff
@@ -26,4 +31,71 @@ type Gateway struct {
...
+			}
+		}()
+	}
+
+	g.updateTimestamp()
+	return g.Status.Update(status)
+}
+
+func (g *Gateway) HandleUplink(uplink *pb_router.UplinkMessage) (err error) {
+	if g.monitor != nil {
```
htdvisser: We probably want to do this after we get the result, because at this point, the func might still fail

core/router/gateway/gateway.go
```diff
@@ -26,4 +31,71 @@ type Gateway struct {
...
+	monitor *monitorConn
+
+	Ctx log.Interface
+}
+
+func (g *Gateway) updateTimestamp() {
+	g.LastSeen = time.Now()
+}
+
+func (g *Gateway) HandleStatus(status *pb.Status) (err error) {
```
htdvisser: We probably want to do this after we get the result, because at this point, the func might still fail

core/router/gateway/monitor.go
```diff
@@ -0,0 +1,101 @@
+package gateway
+
+import (
+	"context"
```
htdvisser: `golang.org/x/net/context`

core/component.go
```diff
@@ -129,6 +130,7 @@ const (
 type Component struct {
 	Identity         *pb_discovery.Announcement
 	Discovery        pb_discovery.Client
+	Noc              pb_noc.MonitorClient
```
rvolosatovs: see 81ec2fa

core/router/server.go
```diff
@@ -21,39 +21,65 @@ type routerRPC struct {
...
+		return errors.BuildGRPCError(err)
+	}
+
+	token, err := tokenFromMetadata(md)
 	if err != nil {
 		return errors.BuildGRPCError(err)
 	}
+
+	//TODO Validate token
+	r.router.getGateway(id).Token = token
```
rvolosatovs: see 	d8142a3

core/router/server.go
rvolosatovs: see 	f6f4409

core/router/gateway/gateway.go
rvolosatovs: see bcdc11d

core/router/gateway/monitor.go
rvolosatovs: see 42496f2

core/router/gateway/gateway.go
rvolosatovs: see bcdc11d

core/router/gateway/gateway.go
rvolosatovs: see bcdc11d

core/router/server.go
rvolosatovs: see f6f4409

core/router/server.go
rvolosatovs: see f6f4409

core/router/server.go
rvolosatovs: see f6f4409

core/router/gateway/gateway.go
```diff
@@ -26,4 +32,87 @@ type Gateway struct {
...
+}
+
+func (g *Gateway) SetToken(token string) {
+	g.tokenLock.Lock()
+	defer g.tokenLock.Unlock()
+
+	g.Token = token
+}
+
+func (g *Gateway) updateTimestamp() {
```
htdvisser: `updateLastSeen`

core/router/gateway_status.go
```diff
@@ -21,11 +20,5 @@ func (r *router) HandleGatewayStatus(gatewayID string, status *pb_gateway.Status
...
 	}()
 
-	gateway := r.getGateway(gatewayID)
-	gateway.LastSeen = time.Now()
-	err = gateway.Status.Update(status)
-	if err != nil {
-		return err
-	}
-	return nil
+	return r.getGateway(gatewayID).HandleStatus(status)
```
htdvisser: @rvolosatovs?

htdvisser:
Is it possible in any way that the gateway token is not set when one of the `connectXXXMonitor` is called?

core/router/gateway_status.go
rvolosatovs: I thought I have addressed that in your previous comment, but it seems that it disappeared.
Please consider https://play.golang.org/p/wMSQ3XbT17 , which I mentioned in my previous reply
Otherwise, I do not exactly understand what the issue is here

rvolosatovs:
no, as long as transferred using GRPC, as the RPC server sets the token

core/router/gateway_status.go
htdvisser: We have special error handling in this func:

```go
  	defer func() {
  		if err != nil {
 			ctx.WithError(err).Warn("Could not handle gateway status")
 		} else {
 			ctx.WithField("Duration", time.Now().Sub(start)).Info("Handled gateway status")
  		}
  	}()
```

For this to work, `err` needs to be set

rvolosatovs: it is set on return, isn't it?
how is this situation different from the one in the playground example?

htdvisser: Didn't work for me earlier. Maybe this changed in a more recent version of go...

core/router/gateway/gateway.go
rvolosatovs: 505da42

core/router/gateway_status.go
rvolosatovs: could be, but also mind that https://play.golang.org/p/j2DP7x2DDU would not work.

rvolosatovs:
please let me know what you think about 891d41c

concerning `gatewayIDFromMetadata`: it seems generic enough to move it to smth like `utils/metadata.GetID`, as I guess not only gateways pass their `id` that way

core/component.go
```diff
@@ -129,6 +130,7 @@ const (
 type Component struct {
 	Identity         *pb_discovery.Announcement
 	Discovery        pb_discovery.Client
+	NOC              pb_noc.MonitorClient
```
rvolosatovs: That was what I initially did, but @htdvisser requested change. will do...

core/component.go
```diff
@@ -13,6 +13,7 @@ import (
 
 	"github.com/TheThingsNetwork/ttn/api"
 	pb_discovery "github.com/TheThingsNetwork/ttn/api/discovery"
+	pb_noc "github.com/TheThingsNetwork/ttn/api/noc"
```
rvolosatovs: so do you mean getting rid of the alias and just importing it as `noc` ?
it seemed that the convention is to import everything from `/api/XXX` as `pb_XXX`

core/router/downlink.go
```diff
@@ -49,12 +49,8 @@ func (r *router) UnsubscribeDownlink(gatewayID string) error {
 	return nil
 }
 
-func (r *router) HandleDownlink(downlink *pb_broker.DownlinkMessage) error {
+func (r *router) HandleDownlink(downlink *pb_broker.DownlinkMessage) (err error) {
```
rvolosatovs: wanted to be consistent with all other functions of `handleXXX`  type(where the variable is used)
are you sure it has to be changed?

core/component.go
rvolosatovs: 2bf2c64

core/router/gateway/gateway.go
```diff
@@ -26,4 +32,87 @@ type Gateway struct {
...
+	if err = g.Status.Update(status); err != nil {
+		return err
+	}
+	g.updateTimestamp()
+
+	if g.monitor != nil {
+		go func() {
+			cl, err := g.statusMonitor()
+			if err != nil {
+				g.Ctx.WithError(errors.FromGRPCError(err)).Error("Failed to establish status connection to the NOC")
```
rvolosatovs: 2bf2c64

core/router/gateway/gateway.go
```diff
@@ -26,4 +32,87 @@ type Gateway struct {
...
+			}
+
+			if err = cl.Send(status); err != nil {
+				g.Ctx.WithError(errors.FromGRPCError(err)).Error("NOC status push failed")
```
rvolosatovs: 2bf2c64

core/component.go
htdvisser: @johanstokking: That is not true, as it is an abbreviation of Network Operations Center. Similar to Internet Protocol being `IP` and with `ID`, `EUI` etc. Unfortunately the generated protos are different, but this is not a Go convention.

core/router/downlink.go
rvolosatovs: f4939d1

htdvisser:
@johanstokking?

PR #271, opened 2016-09-26
Feature/use account lib

romeovs:
Removes the local version of `core/account` in favor of https://github.com/TheThingsNetwork/go-account-lib

Notable changes:
- `tokenkey` provider is now a part of `go-account-lib` and supports different caching strategies
- `ttnctl` now stores tokens and derived tokens in the directory specified by `--token-dir` (default `$HOME/.ttnctl`) instead of in the config file, it names the file after which account server the token is for.


core/discovery/server.go
```diff
@@ -65,7 +65,7 @@ func (d *discoveryServer) checkMetadataEditRights(ctx context.Context, in *pb.Me
 		}
 		// Allow APP_ID announcements from all trusted auth servers
 		// When announcing APP_ID, token is user token that contains apps
-		if !claims.CanEditApp(string(in.Metadata.Value)) {
+		if !claims.AppRight(string(in.Metadata.Value), "settings") {
```
johanstokking: Create a `const` for this

romeovs:
Do not merge yet, I'm resolving an issue with the handler manager in `ttnctl`

romeovs:
This should be ready for a merge

ttnctl/cmd/applications_pf.go
```diff
@@ -33,7 +33,7 @@ function Decoder(bytes) {
 
 		appID := util.GetAppID(ctx)
 
-		conn, manager := util.GetHandlerManager(ctx)
+		conn, manager := util.GetHandlerManager(ctx, appID)
 		defer conn.Close()
 
 		app, err := manager.GetApplication(appID)
```
htdvisser: Why do we need to pass `appID` to `manager.GetApplication` when `manager` is already scoped to this `AppID`? (same for other `ttnctl` cmds)

romeovs: The manager does not actually know what appID its token is for. We could refactor it to know though.

romeovs: Basically:
```
util.GetHandlerManager(ctx, appID)
```
Gets a handler manager that is preloaded with the correct token.
```
app, err := manager.GetApplication(appID)
```
Fetches the application.

PR #274, opened 2016-09-27
api/metadata.go: helper functions for metadata

rvolosatovs:


johanstokking:
@AntoineRondelet this is probably useful for you as well

PR #284, opened 2016-10-03
utils/scramble: scramble package

rvolosatovs:
scramble package provides various scrambling functions.
Is uses [crypto/rand] (https://golang.org/pkg/crypto/rand/) for random number generation.
used, for example, by the NOC - if users token does not grant access to gateway location, it gets scrambled, such that the real location does not get reported.

- [ ] A proper `delta` value should be determined for the gateway location field scrambling. What are your opinions?

### DIFF
--- a/api/gateway/scramble.go
+++ b/api/gateway/scramble.go
@@ -1,0 +1,16 @@
+package gateway
+
+import "github.com/TheThingsNetwork/ttn/utils/scramble"
+
+func (m *GPSMetadata) Scramble() (err error) {
+	if m.Latitude, err = scramble.Float32(m.Latitude, 10+.01*m.Latitude); err != nil {
+		return err
+	}
+	if m.Longitude, err = scramble.Float32(m.Longitude, 10+.01*m.Longitude); err != nil {
+		return err
+	}
+	if m.Altitude, err = scramble.Int32Delta64(m.Altitude, int64(10+m.Altitude/100)); err != nil {
+		return err
+	}
+	return nil
+}
--- a/utils/scramble/scramble.go
+++ b/utils/scramble/scramble.go
@@ -1,0 +1,67 @@
+package scramble
+
+import (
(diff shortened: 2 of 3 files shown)


utils/scramble/scramble.go
```diff
@@ -0,0 +1,67 @@
+package scramble
```
htdvisser: Please add these to `utils/random` package

api/gateway/scramble.go
```diff
@@ -0,0 +1,16 @@
+package gateway
+
+import "github.com/TheThingsNetwork/ttn/utils/scramble"
+
+func (m *GPSMetadata) Scramble() (err error) {
```
htdvisser: Why are gateways in Northeast Russia scrambled over a larger distance than gateways on null island?

htdvisser:
I don't think this code belongs in `TheThingsNetwork/ttn`. As far as I understand the functions you're adding are only needed in the NOC component. Why should they be implemented in `ttn`?

rvolosatovs:
I am not sure actually. Thought it may be useful for `ttn` also.

api/gateway/scramble.go
rvolosatovs: That's why I added the "A proper delta value should be determined for the gateway location field scrambling. What are your opinions?" point.
How, do you think, this should be done?

rvolosatovs:
also, the `gateway.Location.Scramble()` seemed to be convenient and make sense

johanstokking:
This should not be in `ttn`:

1. There is a trust relationship from the Router to its monitors anyway, since it forwards everything. There is no need to scramble anything before that, just trust the NOC to respect public/private status when exposing gateway status (location, counters etc)

2. Convenience code that we don't use should not be part of this codebase. `ttn` shall be lean and clean. Stuff for later use can be in branches, it really necessary

PR #285, opened 2016-10-05
Add newline after options

FokkeZB:
Or the heading follow it won't be styled

/cc @htdvisser


FokkeZB:
Done. BTW... this isn't very DRY, can't it re-use same script?

htdvisser:
Fixed

PR #289, opened 2016-10-06
NOC initialization

rvolosatovs:
Closes #268

PR #296, opened 2016-10-10
Monitor initialization and integration for gateways

### LINKED ISSUE #268: Connection to NOC should be initialized in `ttn`

rvolosatovs:
Closes #268 
Reimplements #261

- Allows to specify multiple monitors
- Reopens the stream in case token has expired(expected behavior in the NOC)

I did not manage to reopen #289, probably connected with the switch to new branch naming scheme.


core/component.go
```diff
@@ -118,6 +118,18 @@ func NewComponent(ctx log.Interface, serviceName string, announcedAddress string
 		go http.ListenAndServe(fmt.Sprintf(":%d", healthPort), nil)
 	}
 
+	if monitors := viper.GetStringMapString("monitor-servers"); len(monitors) != 0 {
+		component.Monitors = map[string]pb_noc.MonitorClient{}
```
johanstokking: Use `make(map[string]pb_noc.MonitorClient)`

api/noc/client.go
```diff
@@ -0,0 +1,17 @@
+package noc
```
johanstokking: Rename to `monitor`. NOC is a product, not a role

core/component.go
htdvisser: And you should probably move it outside of the `if`

core/router/gateway/monitor.go
```diff
@@ -1,102 +1,191 @@
 package gateway
 
 import (
```
htdvisser: Please clean up imports. There should not be an empty line after `google.golang.org/grpc/metadata` nor after `context "golang.org/x/net/context`.

core/router/gateway/monitor.go
```diff
@@ -1,102 +1,191 @@
...
 
-func (g *Gateway) connectUplinkMonitor() (client pb_noc.Monitor_GatewayUplinkClient, err error) {
-	if client, err = connectUplinkMonitor(g.monitorContext(), g.monitor.client); err != nil {
-		return nil, err
-	}
-	g.monitor.uplink.client = client
-	return client, nil
+func NewMonitorConn(clients map[string]pb_noc.MonitorClient) (conn *monitorConn) {
+	conn = &monitorConn{clients: clients}
+	conn.uplink.streams = map[string]pb_noc.Monitor_GatewayUplinkClient{}
```
htdvisser: use `make(map....`

core/router/gateway/monitor.go
```diff
@@ -1,102 +1,191 @@
...
 }
 
-func (g *Gateway) connectDownlinkMonitor() (client pb_noc.Monitor_GatewayDownlinkClient, err error) {
-	if client, err = connectDownlinkMonitor(g.monitorContext(), g.monitor.client); err != nil {
-		return nil, err
+func (g *Gateway) pushStatusToMonitor(ctx log.Interface, name string, status *pb.Status) (err error) {
+	defer func() {
+		switch err {
+		case nil:
+			ctx.Info("Pushed status to monitor")
```
htdvisser: Either make this a `Debug` or remove completely. Don't want to fill server logs with "pushed X to monitor". Same for Uplink and Downlink.

core/router/gateway/monitor.go
```diff
@@ -1,102 +1,191 @@
...
+
+		case io.EOF:
+			ctx.Info("Stream closed, retrying")
+			g.monitor.status.Lock()
+			delete(g.monitor.status.streams, name)
+			g.monitor.status.Unlock()
+			err = g.pushStatusToMonitor(ctx, name, status)
```
htdvisser: If the NOC refuses the status (for example because of an invalid token) this will create an infinite loop. Same for Uplink and Downlink.

core/router/gateway/monitor.go
```diff
@@ -1,102 +1,191 @@
...
-	if client, err = connectDownlinkMonitor(g.monitorContext(), g.monitor.client); err != nil {
-		return nil, err
+func (g *Gateway) pushStatusToMonitor(ctx log.Interface, name string, status *pb.Status) (err error) {
+	defer func() {
+		switch err {
+		case nil:
+			ctx.Info("Pushed status to monitor")
+
+		case io.EOF:
+			ctx.Info("Stream closed, retrying")
```
htdvisser: Warn. Same for Uplink and Downlink.

htdvisser: Also be more specific about what stream was closed exactly.

core/component.go
rvolosatovs: why? we can just check if `component.Monitors` is `nil`, right?

core/router/gateway/monitor.go
rvolosatovs: There is only one stream to be closed and the function is called with the `ctx`, which contains the monitor name
what else should I add?

core/router/gateway/monitor.go
rvolosatovs: not sure about that. if a stream is closed, then it is deleted from the stream hashmap and so it is recreated on the next run(if it was not already created by another goroutine). If the creation fails, then an `err` is returned(it will, if token does not grant access to the gateway), if a stream is opened and an `io.EOF` is returned again, then indeed a loop happens.
The infinite loop is only possible, if token expires at the time moment between opening a stream and sending on it. correct me if I'm wrong...

api/noc/client.go
rvolosatovs: I actually commented about it in #289 , as I was not sure if that's ok to do

core/component.go
rvolosatovs: and I actually thought that the short form(`map[string]]pb_noc.MonitorClient{}`) is easier to read and it has no difference for the compiler, but I will do that as a convention then.

core/router/gateway/monitor.go
htdvisser: Ok then it's fine

core/router/gateway/monitor.go
htdvisser: I think you should just remove the recursion. A missed NOC push is not critical, while a loop might break routing services.

api/noc/client.go
htdvisser: Just do it ;)

core/component.go
htdvisser: I think the convention is to use `make(map[string]...` if you initialize it as empty.
If you check for `nil` it's fine, but for other developers it's really easy to forget this.

rvolosatovs: so what do you want me to do here? I think checking for `nil` goes well together with checking if `gateway.Monitor` is `nil`

htdvisser:
@rvolosatovs: Could you update your branch to include the latest changes in the base branch?
@johanstokking: Do you approve the changes?

PR #308, opened 2016-10-12
Return correct config file, even if none were found

romeovs:
Returns a fallback config location, even if it does not exist. This way people know where it looks.

Also tries both `.ttnctl.yaml` and `.ttnctl.yml` to avoid unnecessary breakage.

### DIFF
--- a/ttnctl/util/config.go
+++ b/ttnctl/util/config.go
@@ -29,22 +29,19 @@ const (
 // $HOME/.ttnctl.yml
 func GetConfigFile() string {
 	flag := viper.GetString("config")
+	xdg := path.Join(os.Getenv("XDG_CONFIG_HOME"), "ttnctl", "config.yml")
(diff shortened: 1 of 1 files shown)


ttnctl/util/config.go
```diff
@@ -29,30 +29,32 @@ const (
 // $HOME/.ttnctl.yml
 func GetConfigFile() string {
 	flag := viper.GetString("config")
+	xdg := path.Join(os.Getenv("XDG_CONFIG_HOME"), "ttnctl", "config.yml")
```
htdvisser: If `XDG_CONFIG_HOME` is unset, this means `xdg` will be `ttnctl/config.yml`, right?
That will lead to an incorrect location in the `try_files` loop.

romeovs: You're right! I fixed it.

PR #315, opened 2016-10-18
Use Redis for storage

htdvisser:
This Pull Request adds a new Redis storage driver. The discovery, networkserver and handler logic has been updated to use this new storage driver. The in-memory storage drivers (previously only used for testing) have been deleted.


htdvisser:
Merged

Issue #203, opened 2016-07-08
Disable persistent sessions in MQTT


johanstokking:
Currently, the MQTT broker allows persistent clients (clients with clean sessions disabled), e.g. they rely on the broker to restore the session with subscriptions. This is not safe as the MQTT broker always needs to remain state of the clients and subscriptions, and also share this state among brokers when we scale it horizontally. When the MQTT broker is unable to restore the session, the MQTT protocol does not provide a way to inform the client. Disconnecting the client is not an option, because of auto reconnect, the client will keep connecting automatically.

- [ ] Do not allow clients with clean session disabled on the MQTT broker **
- [ ] The MQTT broker should not send PUBACK or SUBACK when QoS 2 is requested **
- [x] Update the default MQTT client to enable clean sessions (as part of #199)
- [ ] Update the SDKs and examples to enable clean sessions

Note: this breaks QoS 1 and 2 for offline clients. QoS 1 is still supported when the client is online.

** this may require forking Mosquitto. We may need to consider building our own MQTT broker, which becomes way easier if we don't need to support persistent sessions.

kgbvax:
I think that we will loose too much if we throw all this overboard. 
Where is the point of having a queue when it is not queuing anything? (At least this is how I understand this issue).

Being able to have persitent session and "work" in the eye of failing clients and connections is one of the major benefits of the async model.

I don't get what the problem with mosquitto is but perhaps you want to look at emqttd?
I ran a public broker for a while and persistent session were not a problem as the broker has a high-water mark for the number of messages and duration how long it would retain messages for a single persistent client, preventing it from overflowing.

johanstokking:
Although the Q in MQTT stands for Queuing, it's just a feature (and an old name). MQTT is primarily a publish/subscribe protocol, and that is what we need.

The problem is not so much in the broker, but more in the behavior of the clients when they are persistent.

When clients are persistent, they require the broker to restore the session on reconnect, so they subscribe once on start-up (which may be months ago). That is by design of MQTT. Now, if the broker is unable to restore the session, because of expiration or losing state, it has no way to inform the clients, so they are connected but they are not subscribed, let alone that they will get their lost messages.

If queuing is really necessary, then we might need to use AMQP for that. MQTT is just not the way to go here in this scenario.

kgbvax:
>it has no way to inform the clients, so they are connected but they are not subscribed, let alone that they will get their lost messages.

[...]

PR #319, opened 2016-10-20
Publish uplink over AMQP

johanstokking:
And promoted types from `mqtt` to `core/types`


amqp/publisher.go
```diff
@@ -0,0 +1,79 @@
...
+		err = action()
+	}
+	return err
+}
+
+func (p *DefaultPublisher) publish(key string, msg []byte, timestamp time.Time) error {
+	return p.do(func() error {
+		return p.channel.Publish(p.exchange, key, false, false, AMQP.Publishing{
+			ContentType:  "application/json",
+			DeliveryMode: AMQP.Persistent,
```
htdvisser: This is not consistent with https://github.com/TheThingsNetwork/ttn/issues/203

johanstokking: That's right, but #203 is more related to flaws in MQTT than with persistency in general

amqp/uplink.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/TheThingsNetwork/ttn/core/types"
+)
+
+// PublishUplink publishes an uplink message to the MQTT broker
```
htdvisser: AMQP

mqtt/downlink.go
```diff
@@ -7,14 +7,16 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/TheThingsNetwork/ttn/core/types"
+
```
htdvisser: fmt

amqp/publisher.go
htdvisser: 👍

PR #323, opened 2016-10-23
AMQP downlink

johanstokking:


core/handler/handler.go
```diff
@@ -107,7 +107,7 @@ func (h *handler) Init(c *core.Component) error {
 	}
 
 	if h.amqpEnabled {
-		err = h.HandleAMQP(h.amqpUsername, h.amqpPassword, h.amqpHost, h.amqpExchange)
+		err = h.HandleAMQP(h.amqpUsername, h.amqpPassword, h.amqpHost, h.amqpExchange, "ttn-handler-downlink")
```
htdvisser: Maybe make that a `const`?

PR #324, opened 2016-10-24
monitor integration: #296 reimplementation

rvolosatovs:
- adds api/monitor.Client - wrapper for monitor.MonitorClient
- adds api/monitor.GatewayClient
- uses api/monitor.GatewayClient instead of
  core/router/gateway/monitor.go
- improves logging functionality


api/monitor/client.go
```diff
@@ -1,17 +1,476 @@
 package monitor
 
 import (
+	"context"
```
htdvisser: use x context for now (will change that soon)

Issue #247, opened 2016-09-16
Payload Functions: Pass port as second argument


FokkeZB:
Users will address different types of messages to different ports. In the payload functions I need to know to what port the message was addressed. Currently, I don't have this information.

I'd suggest we add the port as second argument to all 4 functions:

```js
function Decoder(bytes, port) {
  // Decode an uplink message from
  // a Buffer of bytes to an object.

  decodedObj = {};

  if (port === 1) {
    decodedObj.isLightOn = bytes[0];
  }

  return decodedObj;
}
```

Once this is in place here, https://github.com/TheThingsIndustries/dashboard/issues/74 requests for the placeholders to demonstrate usage of the port.

htdvisser:
@johanstokking promised on Slack to also add frame counters to the payload functions (https://thethingsnetwork.slack.com/archives/backend/p1474722721000621). I don't see the added value of adding these counters, especially because they are only relevant for LoRaWAN devices. Could you please elaborate, @johanstokking?

johanstokking:
Yeah we don't need the frame counters indeed, but port is essential.

FokkeZB:
Can we get this port thing rolling? I need it for the Electra demo and Node QS

romeovs:
Closed by #329

FokkeZB:
@romeovs could you create an issue or PR in the docs repo when you make a change that affects users?

https://github.com/TheThingsNetwork/docs/issues/66

FokkeZB:
Shouldn't these be updated as well?

- https://github.com/TheThingsNetwork/ttn/blob/v2-preview/core/handler/convert_fields_test.go#L50
- https://github.com/TheThingsNetwork/ttn/blob/v2-preview/core/handler/dry_run_test.go#L77

PR #329, opened 2016-10-24
Add ports to payload functions and dry payload functions

romeovs:
Adds ports to payload functions as second argument. Closing #247.


htdvisser:
For your tests: Messages with FPort `0` do not contain a payload.

htdvisser:
Please add a test with a payload function that does not take a second argument to verify that Otto accepts this as well

romeovs:
@htdvisser I think most cases are covered in the tests.
It is hard to tell because the tests are a bit messy. If you want I can rewrite the tests to be more general and clear.

About FPort being `0`, I've used different ports throughout the test, not only `0`. It should not matter in these tests however.

PR #330, opened 2016-10-24
Declare exchanges out of band, handle multiple uplink messages

johanstokking:


htdvisser:
Please document how these exchanges should be declared out of band

PR #331, opened 2016-10-25
Feature/payload logs

romeovs:
Adds logging when dry-running payload functions, allowing people to debug more easily.

Also factors the js code running  functionality into a separate package, with it's own tests.


api/handler/handler.proto
```diff
@@ -82,14 +82,21 @@ message DryUplinkMessage {
...
+  string          function = 1;
+  repeated string fields   = 2;
+}
+
 message DryUplinkResult {
-  bytes  payload  = 1;
-  string fields   = 2;
-  bool   valid    = 3;
+  bytes             payload = 1;
+  string            fields  = 2;
```
htdvisser: Please add comment explaining that this is JSON

core/handler/functions/logger.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+// vm is used for stringifying values
+var vm = otto.New()
+
+// JSON stringifies a value inside of the otto vm, yielding better
+// results than Export for Object-like class such as Date, but being much
+// slower.
+func JSON(val otto.Value) string {
```
htdvisser: Don't we need some kind of mutex here?

romeovs: You're right, I just lifted the `vm` out of the function without thinking about that.

PR #332, opened 2016-10-25
Allow for additional code by calling named payload function

### LINKED ISSUE #305: Payload Functions: Allow additional functions called from PF
We currently wrap the payload functions in `(%s)(payload)` when we execute them:
https://github.com/TheThingsNetwork/ttn/blob/v2-preview/core/handler/convert_fields.go#L70

This makes that if you add another function after your function to do something like:

```js
function Decoder(bytes) {
  return {
    foo: bar(bytes)
  };
}

function foo() {
  return 'alpha';
}
```

You get:

```
Error("(anonymous): Line 7:1 Unexpected token function (and 5 more errors)")
```

We could instead wrap the code in:

```js
%s
Decoder(payload);
```

Or go common-JS-ish with:

```js
exports = {};
%s
exports.Decoder(payload);
```

romeovs:
Closes #305 by using @FokkeZB 's proposal.

### DIFF
--- a/core/handler/convert_fields.go
+++ b/core/handler/convert_fields.go
@@ -73,7 +73,10 @@ func (f *UplinkFunctions) Decode(payload []byte, port uint8) (map[string]interfa
 		"payload": payload,
 		"port":    port,
 	}
-	code := fmt.Sprintf("(%s)(payload.slice(0), port)", f.Decoder)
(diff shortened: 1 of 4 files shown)


romeovs:
This is a breaking change, payload functions that are not named will no longer work.

PR #336, opened 2016-11-01
Explicit reopen of channel after reconnect

johanstokking:


amqp/client_test.go
```diff
@@ -95,6 +95,7 @@ func TestReopenChannelClient(t *testing.T) {
 	defer p.Close()
 
 	test := func() error {
+		ctx.Debug("Testing publish")
```
htdvisser: ?

johanstokking: To see what happens when when testing in verbose mode; whether the "Testing publish" is before or after reopening the channel.

PR #344, opened 2016-11-04
Add windows specific info for RabbitMQ

Niondir:
Making things clear which I failed on.

PR #345, opened 2016-11-04
Make everything run on windows

Niondir:
I updated the readme to use  `make dev` instead of `make link`, since linking would fail on windows.

Second change is to replace the `.env/broker/networkserver.cert` symlink by a copy of the certificate.


Niondir:
Okay seems like updating the branch helped. Let me know if I have to clean up the history somehow else.

johanstokking:
While you're at it, please fix the typo introduced in https://github.com/TheThingsNetwork/ttn/pull/344

PR #371, opened 2016-11-29
Use plaintext instead of js(on)

FokkeZB:
Comments in JSON do not render well

### DIFF
--- a/mqtt/README.md
+++ b/mqtt/README.md
@@ -12,7 +12,7 @@
 
 **Message:**
 
-```js
+```plaintext
 {
   "port": 1,                          // LoRaWAN FPort
   "counter": 2,                       // LoRaWAN frame counter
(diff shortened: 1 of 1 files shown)


htdvisser:
For that exact reason, these comments use `js`, not `json`. This renders fine, as you can see in this screenshot:

screen shot 2016-11-29 at 20 19 04

FokkeZB:
Unfortunately not in rouge, what jekyll uses: https://www.thethingsnetwork.org/docs/v2-preview/mqtt/#uplink-messages

htdvisser:
Can't you just configure `redcarpet` and `pygments`?

FokkeZB:
Yes, if that is closer to what Github renders we should indeed. Will give it a try.

FokkeZB:
Created https://github.com/TheThingsNetwork/docs/issues/100

PR #381, opened 2016-12-16
core/proxy.WithPagination

rvolosatovs:
Used in the NOC and most probably useful for ttn in the future


htdvisser:
I'd prefer to have a version with `?limit=20&offset=20` to have more control over how many results are returned.

PR #383, opened 2016-12-16
Extended steps for preparing the development environment with the ins…

jwillemsen:
…tallation of govendor, not normally installed on for example OpenSuSE


htdvisser:
Govendor is installed by `make build-deps`, which should be called when you run `make dev-deps` (see https://github.com/TheThingsNetwork/ttn/blob/master/Makefile#L21-L27). Please verify if you really ran `make dev-deps`.

jwillemsen:
Very strange, can't reproduce it anymore, initially had the error that govendor couldn't be found. Will try to reproduce this later on another machine

PR #389, opened 2016-12-19
Re-use Broker-Handler gRPC connections for Activations

htdvisser:


core/broker/broker.go
```diff
@@ -166,32 +166,70 @@ func (b *broker) getRouter(id string) (chan<- *pb.DownlinkMessage, error) {
...
+	}
+	announcement, err := b.Discover("handler", id)
+	if err != nil {
+		return nil, err
+	}
+	conn, err := announcement.Dial()
+	if err != nil {
+		return nil, err
 	}
-	return nil, errors.NewErrInternal(fmt.Sprintf("Handler %s not active", id))
```
romeovs: `errors.Unavailable` would top this

PR #400, opened 2017-01-09
Add auth/trust for gateways

### LINKED ISSUE #392: Add "authenticated" indicator for gateways
Some gateways will be authenticated, some won't be. We need to add an `authenticated` boolean to gateway statuses and gateway metadata to tell the difference.

htdvisser:
This PR adds a `trusted` indicator for gateways. This indicator is set by the Router if the gateway is authenticated with a valid token.

Resolves #392


core/router/server.go
```diff
@@ -37,26 +37,33 @@ func (r *routerRPC) gatewayFromMetadata(md metadata.MD) (gtw *gateway.Gateway, e
 		return nil, err
 	}
 
+	authErr := errors.NewErrPermissionDenied("Gateway not authenticated")
```
rvolosatovs: shouldn't this be a global `var`?

core/router/server.go
```diff
@@ -37,26 +37,33 @@ func (r *routerRPC) gatewayFromMetadata(md metadata.MD) (gtw *gateway.Gateway, e
...
 		claims, err := claims.FromToken(r.router.TokenKeyProvider, token)
 		if err != nil {
-			return nil, errors.NewErrPermissionDenied(fmt.Sprintf("Gateway token invalid: %s", err.Error()))
-		}
-		if claims.Type != "gateway" || claims.Subject != gatewayID {
-			return nil, errors.NewErrPermissionDenied("Gateway token not consistent")
+			authErr = errors.NewErrPermissionDenied(fmt.Sprintf("Gateway token invalid: %s", err.Error()))
+		} else {
+			if claims.Type != "gateway" || claims.Subject != gatewayID {
+				authErr = errors.NewErrPermissionDenied("Gateway token not consistent")
```
rvolosatovs: shouldn't this be a global `var`?

core/router/server.go
```diff
@@ -37,26 +37,33 @@ func (r *routerRPC) gatewayFromMetadata(md metadata.MD) (gtw *gateway.Gateway, e
...
+	authenticated := false
 	token, _ := api.TokenFromMetadata(md)
 
-	if !viper.GetBool("router.skip-verify-gateway-token") {
-		if token == "" {
-			return nil, errors.NewErrPermissionDenied("No gateway token supplied")
-		}
+	if token != "" {
 		if r.router.TokenKeyProvider == nil {
 			return nil, errors.NewErrInternal("No token provider configured")
```
rvolosatovs: shouldn't this be a global `var`?

core/router/gateway/gateway.go
```diff
@@ -81,6 +90,11 @@ func (g *Gateway) HandleUplink(uplink *pb_router.UplinkMessage) (err error) {
 		}
 	}
 
+	// Inject Authenticated
```
rvolosatovs: see above

core/router/gateway/gateway.go
```diff
@@ -32,14 +33,19 @@ type Gateway struct {
 	Schedule    Schedule
 	LastSeen    time.Time
 
-	token string
+	mu            sync.RWMutex // Protect token and Authenticated
```
rvolosatovs: either _token_ should also start with capital or _Authenticated_ should be lowercase

core/router/server.go
```diff
@@ -37,26 +37,33 @@ func (r *routerRPC) gatewayFromMetadata(md metadata.MD) (gtw *gateway.Gateway, e
...
 		}
 		claims, err := claims.FromToken(r.router.TokenKeyProvider, token)
 		if err != nil {
-			return nil, errors.NewErrPermissionDenied(fmt.Sprintf("Gateway token invalid: %s", err.Error()))
-		}
-		if claims.Type != "gateway" || claims.Subject != gatewayID {
-			return nil, errors.NewErrPermissionDenied("Gateway token not consistent")
+			authErr = errors.NewErrPermissionDenied(fmt.Sprintf("Gateway token invalid: %s", err.Error()))
+		} else {
+			if claims.Type != "gateway" || claims.Subject != gatewayID {
```
rvolosatovs: `claims.FromToken` already checks is `claims.Type != "gateway"`

core/router/server.go
htdvisser: imo, this should panic, but that's outside the scope of this PR

core/router/server.go
htdvisser: could do this, but as we don't re-use it anywhere else it just fragments the code

core/router/server.go
htdvisser: could do this, but as we don't re-use it anywhere else it just fragments the code

htdvisser:
@rvolosatovs: processed your comments

Issue #404, opened 2017-01-10
handler.mqtt-port and handler.amqp-port are announced but don't exist


htdvisser:
Ports should be parsed from the `handler.mqtt-address` and `handler.amqp-address` conf instead

- https://github.com/TheThingsNetwork/ttn/blob/develop/cmd/handler.go#L77-L81
- https://github.com/TheThingsNetwork/ttn/blob/develop/cmd/handler.go#L92-L96

johanstokking:
This is done?

PR #405, opened 2017-01-10
Added parse function to retrieve the port from an address

AntoineRondelet:
see #404

### DIFF
--- a/utils/parse/parse.go
+++ b/utils/parse/parse.go
@@ -1,0 +1,26 @@
+package parse
+
+import (
+	"errors"
+	"strconv"
+	"strings"
+)
+
+// Port returns the port from an address
+func Port(address string) (uint, error) {
+	parts := strings.SplitN(address, ":", 2)
+	if len(parts) != 2 {
+		return 0, errors.New("Could not parse the port: malformated address")
+	}
+
+	port, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return 0, err
+	}
+
+	if port < 0 {
+		return 0, errors.New("Invalid port number")
+	}
+
+	return uint(port), nil
+}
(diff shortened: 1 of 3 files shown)


utils/parse/parse.go
```diff
@@ -0,0 +1,26 @@
+package parse
+
+import (
+	"errors"
+	"strconv"
+	"strings"
+)
+
+// Port returns the port from an address
+func Port(address string) (uint, error) {
```
htdvisser: This doesn't support the `user:pass@host:port` scheme, which we might want to use in the future. Let's be future-proof ;)

AntoineRondelet: right

utils/parse/parse.go
```diff
@@ -8,16 +8,17 @@ import (
 
 // Port returns the port from an address
 func Port(address string) (uint, error) {
-	parts := strings.SplitN(address, ":", 2)
-	if len(parts) != 2 {
+	parts := strings.Split(address, ":")
```
johanstokking: We have a builtin `url.Parse()` for this. This gives a `Host` field which you can safely check on presence of a port (format is always `host:port`)

htdvisser: That only works when you supply a `scheme://`, which is not the case here.

htdvisser: From https://golang.org/pkg/net/url/#URL:

> A URL represents a parsed URL (technically, a URI reference). The general form represented is:
> 
> ```
> scheme://[userinfo@]host/path[?query][#fragment]
> ```
>
> URLs that do not start with a slash after the scheme are interpreted as:
> 
> ```
> scheme:opaque[?query][#fragment]
> ```

That would be incorrect.

PR #407, opened 2017-01-11
Added regex to validation error

AntoineRondelet:
This PR is a proposal. 
Despite the fact that the format of the regex is proper to golang I would like to add the regex to the error message when a validation is not successful. This is pretty useful for the user when he gets a "Invalid format" error message.

### DIFF
--- a/api/validation.go
+++ b/api/validation.go
@@ -4,7 +4,6 @@
 package api
 
 import (
-	"fmt"
 	"reflect"
 	"regexp"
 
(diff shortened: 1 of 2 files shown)


htdvisser:
I'm not sure a "regular user" will understand the regex. If we do this, a proper explanation like "IDs can contain lowercase letters, numbers, dashes and underscores and should have a maximum length of 36" would be better.

AntoineRondelet:
I agree with you. Moreover, the format is not the "regular" format of regex, so that might not be the best idea.

The main point is that it could be noce to provide more information to the user concerning IDs. So instead of saying "Wrong ID", it would be better to return "Wrong ID + explanation" :)

johanstokking:
@AntoineRondelet please rebase

Issue #397, opened 2017-01-06
Restart router and monitor streams when token changes


htdvisser:
When the token supplied to the `GatewayClient` (`monitor`) or `RouterClientForGateway` (`router`) changes (for example, from non-authenticated to authenticated), the streams should be re-created with that token in the metadata.

PR #408, opened 2017-01-12
Fix component startup when a monitor server is down

### LINKED ISSUE #396: Component hangs at startup if monitor server down
If a backend component (e.g. broker) has a monitor server in its configuration that is down (e.g. NOC), the component will be stuck in a loop retrying to connect to the monitor during startup. Since the initial connection to the monitors is made synchronously and before the gRPC server is started, the backend component will remain unavailable the entire time.

As far as I understand, monitors are not part of the primary process of the backend. It seems therefore that components should be able to start independently of whether the configured monitor servers are up or down. 

Not really a big issue, but it does introduce some (unnecessary?) brittleness. Any thoughts on this @johanstokking @htdvisser ?

jankramer:
Previously, monitor clients were synchronously initialized during component startup. As a result, if a monitor server was down during during startup, the component would fail to start. This introduces a registry that encapsulates the initialization of monitor clients and exposes a simple API that allows components to retrieve the monitor clients that are ready at that point in time.

Fixes #396.

### DIFF
--- a/api/monitor/registry.go
+++ b/api/monitor/registry.go
@@ -1,8 +1,12 @@
+// Copyright © 2017 The Things Network
+// Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+
 package monitor
 
 import (
-	"github.com/apex/log"
 	"sync"
+
+	"github.com/apex/log"
 )
 
 // Registry encapsulates dealing with monitor servers that might be down during startup.
--- a/api/monitor/registry.go
+++ b/api/monitor/registry.go
@@ -1,0 +1,108 @@
+package monitor
+
+import (
+	"github.com/apex/log"
+	"sync"
+)
+
+// Registry encapsulates dealing with monitor servers that might be down during startup.
+type Registry interface {
+	// InitClient initializes a new monitor client.
+	InitClient(name string, addr string)
+
+	// BrokerClients returns the list of broker monitor clients that have currently been initialized.
+	BrokerClients() []BrokerClient
+
+	// GatewayClients returns the list of gateway monitor clients that have currently been initialized for a given gateway.
+	GatewayClients(id string) []GatewayClient
+
+	// SetGatewayToken configures a token that's applied to all clients for a gateway, regardless of when it's initialized.
+	SetGatewayToken(id string, token string)
+}
+
+// NewRegistry creates a monitor client registry.
+func NewRegistry(ctx log.Interface) Registry {
+	return &registry{
+		ctx:              ctx,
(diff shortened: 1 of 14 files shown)


vendor/vendor.json
```diff
@@ -555,6 +555,12 @@
 			"revisionTime": "2016-12-13T22:48:10Z"
 		},
 		{
+			"checksumSHA1": "+IGrULOjgHRwCPAk45iKxH3jznQ=",
+			"path": "github.com/smartystreets/assertions/should",
```
htdvisser: Please use the `ShouldXXX` functions in `assertions` package, like we do everywhere else.

htdvisser:
@rvolosatovs: Please review when you have time. Should we include the restart-stream-on-token-change changes to the monitor client here as well? (https://github.com/TheThingsNetwork/ttn/issues/397)

htdvisser:
@jankramer: Please add yourself to the `AUTHORS` file.

jankramer:
One of the builds failed due to a failing test in `api/monitor/client_test.go`. Re-running the build without any changes solved the issue, so it looks like those client tests are a little flaky. Is this a known issue @rvolosatovs @htdvisser?

htdvisser:
Haven't seen that before

api/monitor/registry.go
```diff
@@ -0,0 +1,111 @@
...
+		gwClient := client.GatewayClient(id)
+		gwClient.SetToken(r.gatewayTokens[id])
+		r.gatewayClients[id] = append(r.gatewayClients[id], gwClient)
+	}
+
+	return r.gatewayClients[id]
+}
+
+func (r *registry) SetGatewayToken(id string, token string) {
+	r.Lock()
```
jankramer: True, fine-grained locking would be more correct here. However, a single mutex keeps the code simpler. Also, since brokers and routers are run in separate processes, in practice it doesn't really matter. What do you prefer @htdvisser ?

htdvisser:
@rvolosatovs: Can we also merge this before #397? I think I want to make some changes there before merging.

rvolosatovs:
@htdvisser ok

api/monitor/registry.go
htdvisser: I prefer simplicity, so one mutex is fine.

jankramer:
Can this PR be merged, or is there something else blocking?

htdvisser:
- Rebase on `develop`
- File header in `registry.go`
- Format imports

Will merge as soon as CI passes

PR #418, opened 2017-01-16
Restart monitor streams when token changes

### LINKED ISSUE #397: Restart router and monitor streams when token changes

rvolosatovs:
@htdvisser  because of implementation, test results are unpredictable.
often they fail on https://github.com/TheThingsNetwork/ttn/blob/develop/api/monitor/client_test.go#L81 , for example. As for reason why(as far as I understand): we use the `RecvMsg` to determine if there is an error on stream, but that happens concurrently with the `stream.Send`, so there is no way determine which happens first. If the `Send` happens first, the buffer is not going to be full and there will be no error returned. I implemented a _"dirty"_ workaround by sending 2 messages.
I did not find a way to test if stream is indeed reopened on token change in current implementation of tests

closes #397

### DIFF
--- a/api/monitor/client.go
+++ b/api/monitor/client.go
@@ -156,9 +156,10 @@ func (cl *Client) GatewayClient(id string) (gtwCl GatewayClient) {
 	if !ok {
 		cl.mutex.Lock()
 		gtwCl = &gatewayClient{
-			Ctx:    cl.Ctx.WithField("GatewayID", id),
-			client: cl,
-			id:     id,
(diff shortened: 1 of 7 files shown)


htdvisser:
Will be done with upcoming gRPC stream interceptors (WIP in go-utils)

PR #429, opened 2017-01-23
Recover from panics thrown by otto

romeovs:
Do not rethrow panics from the otto runtime.

### DIFF
--- a/core/handler/functions/functions.go
+++ b/core/handler/functions/functions.go
@@ -13,7 +13,7 @@ import (
 
 var errTimeOutExceeded = errors.NewErrInternal("Code has been running to long")
 
-func RunCode(name, code string, env map[string]interface{}, timeout time.Duration, logger Logger) (otto.Value, error) {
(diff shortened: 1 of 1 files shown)


core/handler/functions/functions.go
```diff
@@ -32,22 +32,18 @@ func RunCode(name, code string, env map[string]interface{}, timeout time.Duratio
...
 
-	var value otto.Value
-	var err error
-
 	start := time.Now()
 
 	defer func() {
 		duration := time.Since(start)
 		if caught := recover(); caught != nil {
+			val = otto.Value{}
```
htdvisser: Is already initialized to Zero value. Are you overwriting?

core/handler/functions/functions.go
```diff
@@ -59,7 +55,8 @@ func RunCode(name, code string, env map[string]interface{}, timeout time.Duratio
 			panic(errTimeOutExceeded)
 		}
 	}()
-	val, err := vm.Run(code)
+
+	val, err = vm.Run(code)
 
 	return val, err
```
htdvisser: Can just `return`

PR #430, opened 2017-01-25
Implement Adaptive Data Rate

### LINKED ISSUE #265: Support adaptive data rate (ADR)
Requires Network Server to register whether the node is mobile or static. ADR applies to static nodes only.

htdvisser:
Resolves #265 

@johanstokking: please review

@tftelkamp: please review changes in the following files:

- `core/band/adr.go`
- `core/networkserver/adr.go`
- `core/networkserver/uplink_mac.go`

Remaining questions for @tftelkamp:

- Should we set the ACK bit on downlink messages?
- What can we do when the request is not accepted (one of the three ACKs in the LinkADRAns is 0)?


core/networkserver/adr.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+	}
+	return max
+}
+
+func (n *networkServer) handleUplinkADR(message *pb_broker.DeduplicatedUplinkMessage, dev *device.Device) error {
+	lorawanUplinkMac := message.GetMessage().GetLorawan().GetMacPayload()
+	lorawanDownlinkMac := message.GetResponseTemplate().GetMessage().GetLorawan().GetMacPayload()
+
+	if lorawanUplinkMac.Adr {
```
johanstokking: Is there a `nil` check on `lorawanUplinkMac` required here?

core/networkserver/adr.go
```diff
@@ -0,0 +1,134 @@
...
+		}); err != nil {
+			n.Ctx.WithError(err).Error("Could not push frame for device")
+		}
+		if dev.ADR.Band == "" {
+			dev.ADR.Band = message.GetProtocolMetadata().GetLorawan().GetRegion().String()
+		}
+		dev.ADR.DataRate = message.GetProtocolMetadata().GetLorawan().GetDataRate()
+		if lorawanUplinkMac.AdrAckReq {
+			dev.ADR.SendReq = true
+			lorawanDownlinkMac.Ack = true
```
johanstokking: `nil` check?

core/networkserver/adr.go
htdvisser: This is checked in `uplink.go`, so if it _is_ nil, it should panic.

core/networkserver/adr.go
htdvisser: This should be initialized in `uplink.go`, so if it _is_ nil, it should panic.

tftelkamp:
Ack bit does not need to be set on downlink messages, unless it's a response to a confirmed up. Any downlink is valid to acknowledge an ADRACKReq, even just an empty message.

Regarding the request not being accepted, my interpretation of the specification is that if any of these bits is 0, the entire command has not been applied (LMiC seems to follow that, I have not verified on the Microchip firmware). In this case there is a misalignment between the capabilities/state of the device, and the network server. I'm not sure what to do in this case. Keep trying to send alternate ADR commands does not sound like a good strategy.

PR #433, opened 2017-01-26
Use SCAN instead of KEYS in Redis

htdvisser:


core/storage/redis_map_store.go
```diff
@@ -106,11 +106,20 @@ func (s *RedisMapStore) List(selector string, options *ListOptions) ([]interface
...
+	var allKeys []string
+	var cursor uint64
+	for {
+		keys, next, err := s.client.Scan(cursor, selector, 0).Result()
+		if err != nil {
+			return nil, err
+		}
+		allKeys = append(allKeys, keys...)
+		cursor = next
+		if cursor == 0 {
```
romeovs: Use ```for i >=  0 {``` instead of this explicit break.

core/storage/redis_set_store.go
```diff
@@ -81,11 +81,20 @@ func (s *RedisSetStore) List(selector string, options *ListOptions) (map[string]
...
+	var allKeys []string
+	var cursor uint64
+	for {
+		keys, next, err := s.client.Scan(cursor, selector, 0).Result()
+		if err != nil {
+			return nil, err
+		}
+		allKeys = append(allKeys, keys...)
+		cursor = next
+		if cursor == 0 {
```
romeovs: See above.

core/storage/redis_set_store.go
```diff
@@ -81,11 +81,20 @@ func (s *RedisSetStore) List(selector string, options *ListOptions) (map[string]
 	if !strings.HasPrefix(selector, s.prefix) {
 		selector = s.prefix + selector
 	}
-	keys, err := s.client.Keys(selector).Result()
-	if err != nil {
-		return nil, err
+	var allKeys []string
+	var cursor uint64
+	for {
```
romeovs: I think you need to lock `RedisSetStore` (and `RedisMapStore`). Otherwise interleaved iterations in this `for` loop could result in both invocations only returning half of the `Scan`ned results.

core/storage/redis_map_store.go
htdvisser: Initally `cursor == 0`, but it's also the end condition.

core/storage/redis_set_store.go
htdvisser: Exactly, see above.

PR #442, opened 2017-01-31
Implement Confirmed Downlink

htdvisser:
Resolves #158

PR #443, opened 2017-01-31
Handle retries for confirmed uplinks

### LINKED ISSUE #272: Handle retries for confirmed uplinks
556c49e4b31d4c287ab6ad371c47f322aaa53144

htdvisser:
Resolves #272 

Remaining question: What should the Handler do with retries?

- Just publish on MQTT (assuming the application checks the `counter` value)
- Publish on MQTT, but indicate that it's a retry (assuming the application will check for this flag)
- Don't publish on MQTT at all

Requires change in #442: now that we know a downlink (ACK) was not received by the node, we should also re-send the downlink message.


johanstokking:
Publish with retry indication. This will depend on the application or integration process

PR #470, opened 2017-02-10
Add Confirmed field to MQTT Uplink [API]

htdvisser:


johanstokking:
Let's keep this consistent with retry

PR #474, opened 2017-02-14
Added a Dial function to use TLS without fallback to unsecure

AntoineRondelet:
It might be useful in some cases to tolerate only secure connections with registered CAs (If the client wants to use only TLS without fallback if the server does not use it).

### DIFF
--- a/api/dial.go
+++ b/api/dial.go
@@ -87,6 +87,12 @@ func Dial(address string) (*grpc.ClientConn, error) {
 	return dial(address, tlsConfig, true)
 }
 
+// DialSafely dials an address with TLS only and fails if the server is not using TLS with a known CA
+func DialSafely(address string) (*grpc.ClientConn, error) {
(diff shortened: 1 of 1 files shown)


api/dial.go
```diff
@@ -87,6 +87,12 @@ func Dial(address string) (*grpc.ClientConn, error) {
 	return dial(address, tlsConfig, true)
 }
 
+// DialSafely dials an address with TLS only and fails if the server is not using TLS with a known CA
+func DialSafely(address string) (*grpc.ClientConn, error) {
```
johanstokking: Make it `DialTLS()`

htdvisser:
The only reason we have this fallback, is that (until 2017-01-27) the public Discovery server did not support TLS. In my opinion we should disable the fallback by default and have non-TLS users enable it explicitly if needed.

htdvisser:
https://github.com/TheThingsNetwork/ttn/issues/483

PR #475, opened 2017-02-14
Fix/safe crypto

romeovs:
- Uses the `crypto/rand` package in lieu of `math/rand` for safer tokens and keys.
- Adds a `FillBytes` helper that fills a byte slice completely without using a buffer and use this in all the places `random.Bytes` was used. This should improve performance.
- Add `random.StrongString` which generates a cryptographically strong random string in the `base64.RawURLEncoding`. It is a lot slower than `random.String`:
```
BenchmarkString-4         	 3000000	       365 ns/op
BenchmarkStrongString-4   	 1000000	      2206 ns/op
```
However, it does not need to acquire the lock in the global `TTNRandom`, which, in the end, might improve performance. I propose we remove `String` in favor of `StrongString` altogether.

### DIFF
--- a/core/handler/activation.go
+++ b/core/handler/activation.go
@@ -188,7 +188,7 @@ func (h *handler) HandleActivation(activation *pb_broker.DeduplicatedDeviceActiv
 		// NOTE: As DevNonces are only 2 bytes, we will start rejecting those before we run out of AppNonces.
 		// It might just take some time to get one we didn't use yet...
 		alreadyUsed = false
-		copy(appNonce[:], random.Bytes(3))
+		random.FillBytes(appNonce[:])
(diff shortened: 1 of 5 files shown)


htdvisser:
Good one. Especially for generating the nonces and ABP keys, we should not use `math/rand`.
However, we don't need the (slower) `crypto/rand` everywhere, so I propose that we split this into two packages with (almost) the same API: `random` and `pseudorandom`. Code that does not require crypto (such as generating random IDs for trace messages) can use the faster `pseudorandom` package, while code that does need crypto (such as generating nonces and keys) uses the `random` package.

htdvisser:
Replaced by #481

PR #497, opened 2017-03-01
api: make limit and offset int64

rvolosatovs:


htdvisser:
Could you explain why we need an `int64` instead of an `int`?

rvolosatovs:
to allow higher ranges than `int`

while working on TheThingsIndustries/noc#106
also `redis` ordered sets allow int64 ranges

PR #498, opened 2017-03-02
Implement key exchange using Component Token

htdvisser:
We no longer require client ID + client secret, but allow key->token exchange using component tokens

### DIFF
--- a/core/component/auth_cache.go
+++ b/core/component/auth_cache.go
@@ -1,0 +1,48 @@
+// Copyright © 2016 The Things Network
+// Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+
+package component
+
+import (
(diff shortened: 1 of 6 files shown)


core/component/auth_cache.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/TheThingsNetwork/go-account-lib/cache"
+	"golang.org/x/oauth2"
+)
+
+func authCacheKey(appID, accessKey string) string {
+	return appID + ":" + accessKey
```
romeovs: Consider hashing the access key as it has an arbitrary length. Note that you will need to validate against collisions then however.

PR #522, opened 2017-03-15
Remove `-v` flag from govendor sync

joscha:
`govendor sync` does not have a flag `-v`:

```
govendor sync
	Ensures the contents of the vendor folder matches the vendor file.
	Options:
		-n           dry run, print out action only
		-insecure    allow downloading over insecure connection
```

and thus fails with:
```
➜  ttn git:(develop) make dev-deps
govendor sync -v
Error: flag provided but not defined: -v
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ build-deps:
 	@command -v govendor > /dev/null || go get "github.com/kardianos/govendor"
 
 deps: build-deps
-	govendor sync -v
+	govendor sync
 
 dev-deps: deps
 	@command -v protoc-gen-gogofast > /dev/null || go get github.com/gogo/protobuf/protoc-gen-gogofast


htdvisser:
```
❯ govendor sync --help
govendor sync
	Ensures the contents of the vendor folder matches the vendor file.
	Options:
		-n           dry run, print out action only
		-insecure    allow downloading over insecure connection
		-v           verbose output
```

Please update govendor.

joscha:
It's the latest one available via homebrew - is there a different way to
install it so we'd make it available before it is used in `make`?

htdvisser:
Just uninstall the Homebrew `govendor`. We don't really like working with outdated packages

joscha:
Sure, the Brewfile in the repository kind of suggested brew is a good way
to get started, that's why I used it.

htdvisser:
`govendor` is a Go package, so we use the Go package manager instead of Homebrew.

PR #521, opened 2017-03-15
Add govendor to deps

joscha:
`govendor` is needed to properly install, e.g.:
```
➜  ttn git:(develop) make dev-deps
govendor sync -v
bash: govendor: command not found
make: *** [deps] Error 127
```

PR #524, opened 2017-03-15
use bin path for govendor

joscha:
Fixes #521

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ build-deps:
 	@command -v govendor > /dev/null || go get "github.com/kardianos/govendor"
 
 deps: build-deps
-	govendor sync -v
+	$(GO_PATH)/bin/govendor sync -v
 
 dev-deps: deps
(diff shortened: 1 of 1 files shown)


rvolosatovs:
your env variable name is wrong - it should be `GOPATH`, not `GO_PATH` (that is probably where your issue stems from)

joscha:
That var is defined like that in the Makefile, it has nothing to do with my system. But after two pull requests trying to solve this that have been closed, maybe it's time to move on...

joscha:
For reference: https://github.com/TheThingsNetwork/ttn/blob/develop/Makefile#L9

Sypheos:
govendor is not a part of the standard go library or tools it have to be manually installed on your system.
go get -u github.com/kardianos/govendor

The variable GO_PATH in the Makefile should not be used to access the real GOPATH as it's a modified version to fit our compilation need. Use $GOPATH insteadif you need to.

Also don't forget to set $GOBIN in your system or your shell ans add it to your $PATH variable

PR #542, opened 2017-04-01
Random broker message generators, proper ID formatting

rvolosatovs:
- adds random broker message generators
- fixes bug where `rand.Bool` always returns `true`
- makes sure random `id` generators produce correct `id`s


api/broker/random.go
```diff
@@ -0,0 +1,137 @@
+// Copyright © 2017 The Things Network
+// Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+package broker
```
htdvisser: add empty line before `package broker`

Issue #519, opened 2017-03-14
Pluggable raw payload handling


johanstokking:
This is a **feature request** for **the Backend/ttnctl/the Console**.

Currently, we have the payload functions using a JavaScript sandbox.

Alternatively, we need support for other formats. For example, the user would select `TheThingsMessage` (protobuf), `CayenneLPP`, `Custom (JavaScript)`, etc, where the latter is the current implementation.

This should be as dynamic as possible, potentially leveraging Go 1.8's plugin model. This also makes it easier for the community to add payload handlers without building the back-end from source.

htdvisser:
- Can't these extra formats be implemented as "packaged" JavaScript functions?
- How is using Go 1.8's plugin model easier than building the backend from source?
- How would these "payload processor plugins" be plugged in, let's say, a community Handler?

johanstokking:
Not all formats can be implemented that way, at least in a feasible manner. An example is protobufs.

A plugin model would allow using binaries, whether or not built from source, but at least being able to discover and load formats without (re)building the backend.

In the public community network's Handler, these plugins would be loaded if they're contributed to this repository.

htdvisser:
For the record: I'm against using plugins (at least, for now). Plugins are currently only supported on Linux, and would therefore complicate development and restrict us in deployment options.

htdvisser:
This will be supported in v3 through the `FORMATTER_GRPC_SERVICE`

PR #545, opened 2017-04-05
Payload format in Handler, CayenneLPP support

johanstokking:
This pull request adds an explicit payload format to the application registry in the Handler. Formerly, only JavaScript payload functions are supported, now referred to as `custom` payload format. Now, we can support builtin payload formats, for example CayenneLPP that's included in this pull request (`cayennelpp`).

This sets the base for other builtin payload formats, payload formats that require native code (e.g. protobufs) or even a `device:<manufacturer>:<model>` payload format that refers to a specific device's binary payload.

Addresses #519


core/storage/redis_map_migrate.go
```diff
@@ -19,7 +19,7 @@ type hasDBVersion interface {
 }
 
 // MigrateFunction migrates data from its old version to the latest
-type MigrateFunction func(client *redis.Client, key string, input map[string]string) (version string, output map[string]string, err error)
+type MigrateFunction func(client *redis.Client, key string, input map[string]string) (nextVersion string, output map[string]string, err error)
```
htdvisser: ok

core/handler/application/migrate/2_6_1_payload_format.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+		if encoder, ok := obj["encoder"]; ok {
+			delete(obj, "encoder")
+			obj["custom_encoder"] = encoder
+			any = true
+		}
+		if any {
+			obj["payload_format"] = "custom"
+		}
+		return "2.6.2", obj, nil
```
htdvisser: Should be `2.6.1`. I will change this

PR #552, opened 2017-04-11
v2.6.1

htdvisser:
The 2.6.1 release contains the following new features:

- Support for CayenneLPP payload formats (needs DB migration)

API Changes:

- Payload fields and events are now also published to AMQP

Other changes:

- Moved (memory) debugging and profiling to non-production builds
- Allow uplink-only streams in router client
- Fixed memory alignment and other issues for 32 bit machines


johanstokking:
Whoop whoop

Issue #425, opened 2017-01-20
Register new devices on join


johanstokking:
This is a **feature request** for **the Backend**.

Currently, only devices that are registered can be accepted when joining the network. This requires all devices to be known, including their DevEUI. This might not be practical in large scale deployments.

Therefore, it should be possible for devices to register on join. To avoid join challenges to be sent to all Handlers that registered the AppEUI with a Broker, there should be a first-come-first-serve registration of AppEUI. We might leverage the current unique registration of AppEUI/DevEUI but then with a blank DevEUI. If the Broker doesn't find a specific AppEUI/DevEUI combination for the join, it can challenge a Handler that has the AppEUI registered with a DevEUI as catchall.

## On Security

When allowing any DevEUI to join for an AppEUI and the AppKey leaks, we basically enable spoofing of existing devices and registering new devices. This means that the application needs its own mechanism to identify the device. Many IoT platforms already support this with a device registry. Also, when introducing this feature, we should inform the application owners of the potential risk.

Also, we should restrict the AppEUIs that can register devices anonymously to a range in the Foundation's EUI block. Otherwise, people can abuse The Things Network to send conflicting join accepts for any join request as AppEUIs are public. We can configure the Brokers with an environment variable for a range.

robouden:
Happy to see the request. 
I hope it can be implemented soon.

rob

dotpy3:
This is a proposal for an implementation of this:
+ **User opt-in activation:** The user activates the option through the console, with `ttnctl`, or directly with the API.
  + The user instructs the handler to allow on-join registration for a specific `AppEUI`. The handler stores this decision in the app (through db storage, in the current implementation). An `AppKey` called `AppKey1` is either generated for those devices, or it is specified by the user.
  + The handler sends a new device registration request to its broker `B1`, for an app with a blank `DevEUI`.
  + The broker then retransmits this to the Network Server `NS1` for storage.

+ **First join request scenario:** Considering a new device, with `DevEUI` and `AppEUI`.
  + When configuring the device, the user fills in the `AppEUI` with the right `AppKey`.
  + Router transmits to brokers the activation request, with the device's correct `DevEUI` and `AppEUI`.
[...]

PR #553, opened 2017-04-11
On-join registration

dotpy3:
This pull request solves issue #425.

With the new features of this pull requests, it is now possible for a device to be registered **on-join**, rather than **pre-join**, as it is currently the case.

As of today, if a device wants to send data through TTN, it needs to be registered by an application developer in the console, or with ttnctl. With this new feature, an application developer can specify for an application if s.he wants devices that haven't been registered yet to be registered on-join. Given that a device authenticates by its AppEUI and by its AppKey, the application developer can either specify an AppEUI and/or an AppKey, or let the interface attribute new ones.

# Practical workflow

On-join registration is currently not available in the console.

## Activating on-join registration

Once a user has selected an application in ttnctl, it is possible to enable on-join registration for this app. The AppEUI and AppKey are either specified in the command (`ttnctl applications onjoin-registration --help`), or they can be generated.

```
$ ttnctl applications onjoin-registration on 
  INFO Discovering Handler...                   Handler=dev
  INFO Connecting with Handler...               Handler=localhost:1904
  INFO Generating handler-stored access key    
  INFO Configuring On-Join Registration setting... AppEUI=8000000000000009 AppKey=1E194746761E9923B221B23886D17042 Setting=on
  INFO On-Join Registration setting configured  AppEUI=8000000000000009 AppKey=1E194746761E9923B221B23886D17042 Setting=on
```

In a similar way, it is possible de disable it:

```
$ ttnctl applications onjoin-registration off
  INFO Discovering Handler...                   Handler=dev
  INFO Connecting with Handler...               Handler=localhost:1904
  INFO Configuring On-Join Registration setting... Setting=off
  INFO On-Join Registration setting configured  Setting=off
  INFO Removing access key                      AccessKeyName=on-join-registration-handler-access-key-hello-world-WCQQENCbyr3jNttfwibKQYrVFY0AL7 Setting=off
```

## Using a device

From then, the only necessary thing to do is to specify the right `AppEUI` and `AppKey`, when using a compliant library. It is not necessary to re-generate new AppEUIs and AppKeys every time - they can be reused as long as on-join registration is active.

# Technical implementation

## Activation

+ The request goes from **ttnctl** to the **Handler**, with a new RPC endpoint. The handler reads the request, sets a flag in its internal store, then transmits the request to the broker, that transmits to the **network server**.
+ If the request is considered correct, the network server **saves a new device** in its store, that has no DevID or DevEUI value. This value will be used as a flag that on-join registration is enabled for this AppEUI.

## On-join registration

+ When an activation request arrives at the level of the network server, the network server obviously doesn't find the device. However, another check is now made to see if the on-join registration flag has been set. If it is found, the NS returns to the broker the information that *no similar device has been set, but on-join registration is possible*.
+ The broker, with this information, sends a registration request to the handler. This handler will then send another request to the broker, but with an access token, that has been stored in the handler at the activation of the option. **This is the device registration request.** This allows us to keep this kind of request protected. The request that the handler sends is similar to the one made when a user creates a devices from the console or from the `ttnctl` - allowing us to keep procedure consistency.
+ When the handler has completed the device registration request, it sends the result over to the broker, that can continue the activation request as usual if the request was successful.

## Deactivation

The deactivation process follows the same pattern as activation, only with a different message. The network server deletes the flag from its store, and the handler stores this new information.

# Future works

+ rayozzie, from our community, has illustrated a few applications-level additions to this in the issue: #425. While they are out of scope with the LoRaWAN specifications, they are nevertheless interesting for developers who intent on using this option on a production level.
+ It would make sense to add flags for components to control this option at runtime: for example, limit the number of devices that are allowed per app/per developer to use this option.


johanstokking:
Should the Broker know about this setting in the first place?

I would suggest that the Broker has a AppEUI/blank DevEUI registration, and that it falls back to this registration when the joining AppEUI/DevEUI is not found.

When a Handler sees the blank DevEUI activating, it should use its access key to register a the device on the Broker. So during the normal activation procedure for the Broker, the Broker handles the registration process. Maybe I'm overlooking something, but in this way, we don't need to change the Broker at all and everything is handled by the Handler.

dotpy3:
With this new commit, the messages workflow changes: when an unknown device requests activation by the broker - and the network server indicates that on-join registration is possible, the broker just transmits the request to the handler, that has then the responsability of registering it before pursuing the activation process.

@johanstokking You also talked about pre-registering the device in the network server store so that the handler doesn't have to send messages up to the NS to complete registration. We'd have to be careful of inconsistencies, but it should be doable if you want.

api/handler/handler.proto
```diff
@@ -68,9 +68,15 @@ message Application {
   // message is dropped. This function is used when the payload format is set to custom.
   string validator      = 4;
 
-  // The encoder is a JavaScript function that encodes an object to a byte array.
```
johanstokking: What's wrong with this comment?

api/protocol/lorawan/device.proto
```diff
@@ -53,8 +53,26 @@ message Device {
...
+// SetRegisterOnJoinMessage signals whether we expect on-join registration for an app
+message SetRegisterOnJoinMessage {
+  // App identifier
+  string app_id  = 1;
+  // Toggle on-join registration
+  bool   val     = 2;
+  // AppEui sets the AppEUI non-registered devices have to join with
+  bytes app_eui = 3 [(gogoproto.customtype) = "github.com/TheThingsNetwork/ttn/core/types.AppEUI"];
+  // AppKey sets the AppKey that the new devices will register with 
+  bytes app_key = 4 [(gogoproto.customtype) = "github.com/TheThingsNetwork/ttn/core/types.AppKey"];
```
johanstokking: AppKey always stays in the Handler

api/protocol/lorawan/device.proto
```diff
@@ -53,8 +53,26 @@ message Device {
   int64  last_seen = 21;
 }
 
+// SetRegisterOnJoinMessage signals whether we expect on-join registration for an app
+message SetRegisterOnJoinMessage {
+  // App identifier
+  string app_id  = 1;
+  // Toggle on-join registration
+  bool   val     = 2;
```
johanstokking: `enabled`

api/protocol/lorawan/device.proto
```diff
@@ -53,8 +53,26 @@ message Device {
...
+  // AppEui sets the AppEUI non-registered devices have to join with
+  bytes app_eui = 3 [(gogoproto.customtype) = "github.com/TheThingsNetwork/ttn/core/types.AppEUI"];
+  // AppKey sets the AppKey that the new devices will register with 
+  bytes app_key = 4 [(gogoproto.customtype) = "github.com/TheThingsNetwork/ttn/core/types.AppKey"];
+  // AccessKey to the account server
+  string access_key = 5;
+  // AccessKeyName for the account server key
+  string access_key_name = 6;
```
johanstokking: Broker doesn't need to know this

core/handler/activation.go
```diff
@@ -34,7 +39,72 @@ func (h *handler) getActivationMetadata(ctx ttnlog.Interface, activation *pb_bro
 	return appUp.Metadata, nil
 }
 
+// registerDeviceAtActivation serves the purpose of registering a new device that isn't registered yet,
+// but that has the right IDs to be registered
+func (h *handler) registerDeviceAtActivation(challenge *pb_broker.ActivationChallengeRequest) (*pb_broker.ActivationChallengeRequest, error) {
```
johanstokking: Register device at activation or register on join? Consistent naming

core/handler/manager_server.go
```diff
@@ -388,6 +442,9 @@ func (h *handlerManager) GetApplication(ctx context.Context, in *pb.ApplicationI
 		Converter:     app.CustomConverter,
 		Validator:     app.CustomValidator,
 		Encoder:       app.CustomEncoder,
+		// On-join registration settings
+		OnJoinRegistration:              app.OnJoinRegistration,
+		OnJoinRegistrationAccessKeyName: app.OnJoinRegistrationAccessKeyName,
```
johanstokking: Remove comment; the code below is very clear

core/handler/manager_server.go
```diff
@@ -73,6 +73,60 @@ func (h *handlerManager) validateTTNAuthAppContext(ctx context.Context, appID st
...
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = h.deviceManager.SetRegisterOnJoin(ctx, in)
+	if err != nil {
+		// Try to revert the changes
+		app.OnJoinRegistration = !in.Val
+		if err := h.handler.applications.Set(app); err != nil {
+			h.handler.Ctx.WithField("AppID", app.AppID).Error("Failed to revert the changes at the handler level when SetRegisterOnJoin failed")
```
johanstokking: `WithError(err)` + shorten the log message

core/handler/manager_server.go
```diff
@@ -73,6 +73,60 @@ func (h *handlerManager) validateTTNAuthAppContext(ctx context.Context, appID st
...
+		app.OnJoinRegistrationAppEui = *in.AppEui
+		app.OnJoinRegistrationAppKey = *in.AppKey
+		app.OnJoinRegistrationAccessKey = in.AccessKey
+		app.OnJoinRegistrationAccessKeyName = in.AccessKeyName
+	} else {
+		in.AppEui = &app.OnJoinRegistrationAppEui
+	}
+	h.handler.Ctx.WithFields(ttnlog.Fields{
+		"AppEui":        app.OnJoinRegistrationAppEui,
+		"AppKey":        app.OnJoinRegistrationAppKey,
```
johanstokking: Never log security keys to log

core/handler/manager_server.go
```diff
@@ -73,6 +73,60 @@ func (h *handlerManager) validateTTNAuthAppContext(ctx context.Context, appID st
...
+		"AccessKey":     app.OnJoinRegistrationAccessKey,
+		"AccessKeyName": app.OnJoinRegistrationAccessKeyName,
+	}).Debug("Registering on-join parameters...")
```
johanstokking: Remove dots

core/handler/manager_server.go
johanstokking: Also it might be better to set it locally only when it has been set successfully in the Broker

core/networkserver/manager_server.go
```diff
@@ -48,6 +75,17 @@ func (n *networkServerManager) getDevice(ctx context.Context, in *pb_lorawan.Dev
...
 		return nil, err
 	}
+	if dev == nil {
+		// If no device is found: NS tries to find a device with an empty DevEUI
+		dev, err = n.networkServer.devices.Get(*in.AppEui, emptyDevEUI)
+		if err != nil {
+			return nil, err
+		}
+		if dev == nil {
+			// No empty device either => throwing No Device Found error
```
johanstokking: Remove comment

ttnctl/cmd/applications_onjoinregistration.go
```diff
@@ -0,0 +1,194 @@
...
+		var onJoinSetting bool
+		settingStr := args[0]
+		if settingStr == "on" {
+			onJoinSetting = true
+		} else if settingStr == "off" {
+			onJoinSetting = false
+		} else {
+			// Incorrect string
+			ctx.Fatal("Please choose \"on\" or \"off\" for the On-Join Registration setting.")
+		}
```
johanstokking: Use `switch` and check the lower case of the input

ttnctl/cmd/applications_onjoinregistration.go
```diff
@@ -0,0 +1,194 @@
...
+
+		var onJoinSetting bool
+		settingStr := args[0]
+		if settingStr == "on" {
+			onJoinSetting = true
+		} else if settingStr == "off" {
+			onJoinSetting = false
+		} else {
+			// Incorrect string
+			ctx.Fatal("Please choose \"on\" or \"off\" for the On-Join Registration setting.")
```
johanstokking: Remove caps

ttnctl/cmd/applications_onjoinregistration.go
```diff
@@ -0,0 +1,194 @@
...
+		)
+		assertArgsLength(cmd, args, 1, 3)
+
+		appID := util.GetAppID(ctx)
+		account := util.GetAccount(ctx)
+
+		conn, manager := util.GetHandlerManager(ctx, appID)
+		defer conn.Close()
+
+		var onJoinSetting bool
```
johanstokking: `enabled`

ttnctl/cmd/applications_onjoinregistration.go
```diff
@@ -0,0 +1,194 @@
...
+				random.FillBytes(appKey[:])
+			}
+			bytesKey, err := appKey.MarshalText()
+			if err == nil {
+				resultCtx = resultCtx.WithField("AppKey", string(bytesKey[:]))
+			}
+		} else {
+			accessKeyName = app.OnJoinRegistrationAccessKeyName
+			defer func() {
+				if success {
```
johanstokking: If successful, yeah?

ttnctl/cmd/applications_onjoinregistration.go
```diff
@@ -0,0 +1,194 @@
...
+					resultCtx.WithField("AccessKeyName", accessKeyName).Info("Removing access key")
+					if err := account.RemoveAccessKey(appID, accessKeyName); err != nil {
+						resultCtx.WithError(err).WithField("AccessKeyName", accessKeyName).Error("Failed to remove on-join registration specific access key")
+					}
+				}
+			}()
+		}
+
+		resultCtx.Info("Configuring On-Join Registration setting...")
```
johanstokking: Remove caps

ttnctl/cmd/applications_onjoinregistration.go
```diff
@@ -0,0 +1,194 @@
...
+		err = manager.SetRegisterOnJoin(&lorawan.SetRegisterOnJoinMessage{
+			AppId:         appID,
+			Val:           onJoinSetting,
+			AppEui:        eui,
+			AppKey:        appKey,
+			AccessKey:     accessKey.Key,
+			AccessKeyName: accessKeyName,
+		})
+		if err != nil {
+			resultCtx.WithError(err).Fatal("Could not configure On-Join Registration setting")
```
johanstokking: Idem

ttnctl/cmd/applications_onjoinregistration.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+
+		resultCtx.Info("On-Join Registration setting configured")
```
johanstokking: Idem

api/handler/handler.proto
```diff
@@ -68,9 +68,15 @@ message Application {
...
   // This function is used when the payload format is set to custom.
   string encoder        = 5;
+
+  // Flag that is set if on-join registration is activated on that application.
+  bool   on_join_registration = 7;
+
+  // If on-join registration is activated, contains the name of the access key to interact
+  // with the account server.
+  string on_join_registration_access_key_name = 8;
```
johanstokking: The broker doesn't need to know about this

api/handler/handler.proto
```diff
@@ -68,9 +68,15 @@ message Application {
   // message is dropped. This function is used when the payload format is set to custom.
   string validator      = 4;
 
-  // The encoder is a JavaScript function that encodes an object to a byte array.
   // This function is used when the payload format is set to custom.
   string encoder        = 5;
+
+  // Flag that is set if on-join registration is activated on that application.
+  bool   on_join_registration = 7;
```
johanstokking: Indicate that it's _enabled_ or not

api/protocol/lorawan/device.proto
```diff
@@ -53,8 +53,26 @@ message Device {
...
+}
+
 service DeviceManager {
   rpc GetDevice(DeviceIdentifier) returns (Device);
   rpc SetDevice(Device) returns (google.protobuf.Empty);
   rpc DeleteDevice(DeviceIdentifier) returns (google.protobuf.Empty);
+  // SetRegisterOnJoin activates or deactivates on-join registration for an app
```
johanstokking: Remove comment

api/protocol/lorawan/device.proto
dotpy3: Okay. For the moment, the handler uses the same message format (since it implements the same `DeviceManagerServer` interface), but what I added is that I remove the keys before sending them from the handler. In `core/handler/manager_server.go`, from line 106:
```go
brokerMessage := *in
brokerMessage.AccessKey = ""
brokerMessage.AppKey = &types.AppKey{}
[...]
_, err = h.deviceManager.SetRegisterOnJoin(ctx, &brokerMessage)
if err != nil {
	return nil, errors.New("Broker did not set on-join registration setting")
}
```

ttnctl/cmd/applications_onjoinregistration.go
dotpy3: If on-join registration deactivation was successful on the handler side, `ttnctl` removes the access key. If it wasn't successful, we keep the access key active, so that we don't end up with an on-join registration that is considered enabled, but is actually broken.

api/protocol/lorawan/device.proto
johanstokking: OK. What's advised here @htdvisser ?

htdvisser: Why not put it in the application?

dotpy3: Are you referring to the access key/app key?

johanstokking:
OK @htdvisser can you look?

PR #556, opened 2017-04-12
Adapt for change in GRPC package

rvolosatovs:
Quick fix for https://github.com/grpc/grpc-go/commit/0c1d39d that should not break the existing code


htdvisser:
Let's not do quick and dirty fixes, but keep the vendors as they are, and fix it the right way

PR #558, opened 2017-04-19
Update readme, adding govendor install instruction

Sypheos:
Adding instruction to install govendor.


htdvisser:
`govendor` is installed by `make dev-deps`, which is step 4 of the "Set up The Things Network's backend for Development"

PR #565, opened 2017-04-20
Updated default certificates

dotpy3:
Updated default TLS certificates so that they still work in environments where a component's hostname is different from `localhost`, such as with Docker Compose. For example, for `handler`, this command was executed: `ttn handler gen-cert localhost handler --key-dir ".env/handler"`.


htdvisser:
Found a small issue while reviewing this, so fixed this in a separate commit

PR #585, opened 2017-05-02
Register devices on join

### LINKED ISSUE #425: Register new devices on join

htdvisser:
This PR adds functionality for registering devices when they join the network.

The steps to do this are as follows:

1. `ttnctl applications select [AppID] [AppEUI]`
2. `ttnctl devices register on-join [Device ID Prefix] [AppKey]`
    - This will add an access key to your application if needed
3. Join a device with AppEUI and AppKey - Any DevEUI is accepted
4. The device is registered, its DevID will be `Device ID Prefix`-`DevEUI`

Closes #425


core/handler/activation.go
```diff
@@ -236,15 +249,54 @@ func (h *handler) HandleActivation(activation *pb_broker.DeduplicatedDeviceActiv
...
 	}
 
 	return res, nil
 }
+
+func (h *handler) registerDeviceOnJoin(base *device.Device, activation *pb_broker.DeduplicatedDeviceActivationRequest) (*device.Device, error) {
+	clone := base.Clone()
+	clone.DevID = strings.ToLower(fmt.Sprintf("%s-%s", base.DevID, activation.DevEui.String()))
+	clone.DevEUI = *activation.DevEui
+	clone.Description = fmt.Sprintf("On-Join device registered on %s", time.Now().UTC().Format(time.RFC3339))
```
johanstokking: "Registered on join on %s"

Also can we have a more human readable date/time format here? It doesn't have to be parsable; we have created and updated time stamps

core/handler/activation.go
```diff
@@ -236,15 +249,54 @@ func (h *handler) HandleActivation(activation *pb_broker.DeduplicatedDeviceActiv
...
+
+	app, err := h.applications.Get(base.AppID)
+	if err != nil {
+		return nil, errors.Wrap(err, "Application not registered to this Handler")
+	}
+
+	if app.RegisterOnJoinAccessKey == "" {
+		return nil, errors.NewErrInvalidArgument("Application", "Does not have RegisterOnJoinAccessKey")
```
johanstokking: Does this error propagate to the user? If so, make it human friendly

core/handler/activation_test.go
```diff
@@ -69,105 +78,182 @@ func TestHandleActivation(t *testing.T) {
...
-	var wg WaitGroup
-
-	appEUI := types.AppEUI{1, 2, 3, 4, 5, 6, 7, 8}
-	appID := appEUI.String()
-	devEUI := types.DevEUI{1, 2, 3, 4, 5, 6, 7, 8}
-	devID := devEUI.String()
-	unknownDevEUI := types.DevEUI{8, 7, 6, 5, 4, 3, 2, 1}
 
-	appKey := [16]byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16}
+	// var wg WaitGroup
```
johanstokking: What's this?

ttnctl/cmd/devices_register_on_join.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/TheThingsNetwork/ttn/api/protocol/lorawan"
+	"github.com/TheThingsNetwork/ttn/core/types"
+	"github.com/TheThingsNetwork/ttn/ttnctl/util"
+	"github.com/spf13/cobra"
+)
+
+const registerOnJoinAccessKeyName = "register-on-join"
+
+var devicesRegisterOnJoinCmd = &cobra.Command{
+	Use:   "on-join [Device ID] [AppKey]",
```
johanstokking: Make clear that this is a prefix. Also, be consistent with casing; `DeviceID` or `DevID` (whatever is used elsewhere)

johanstokking:
Looping in @dotpy3 for sanity checking

ttnctl/cmd/devices_register_on_join.go
htdvisser: Will change this to "Device ID Prefix". Taking care of consistency across ttnctl should be done in a separate "refactor" PR

dotpy3:
Overall LGTM, but shouldn't we offer a command in `ttnctl` to disable on-join registration, such as `ttnctl on-join off`? I don't know if it will be clear for users that on-join registration is disabled by removing the template device. Or, we could add an additional info line when enabling on-join registration, such as "Registered on-join device template - remove this device to disable on-join registration". Do the users need to know how on-join registration works within TTN?

htdvisser:
I think the documentation would be a better place to explain this. People who use _and_ ttnctl _and_ on-join registration already know what they're doing.

Issue #507, opened 2017-03-06
Allow key/value pairs as sensor/node metadata


dakoller:
This is a feature request for the Backend.

## Explain what you want to do

When setting up larger usecase, you might end up with different sensors used (which send into the same application), where you want to store e.g. some information about how they are deployed. (e.g. in which height, indoor/outdoor, sensor model, software version of the node)

This information should available to everybody, who listens to the event stream (console, mqtt...), and different people might - based on these metadata - do different calculations with the data.

This key/value dictionary should be read & writeable via API. Additionally each data record, which is outputted via console, mqtt etc. should have the key/value dictionary as additonal information.

johanstokking:
Device attributes as a key/value dictionary can indeed be useful.

A challenge here is to distinguish between generic attributes and custom attributes. The generic attributes are in fact the ones you mention; location, placement, manufacturer, model, version.

There are three possible implementations; 1) adding these as builtin device attributes, 2) introduce key/value attributes with reserved keys and custom keys, or 3) introduce key/value attributes with only custom keys (e.g. don't add UI/UX support for the fields you mentioned in TTN).

I would suggest option 2.

Sypheos:
Option 1 seem more fitting to me.

Though it will be useful, this data should be kept to the minimum.
Adding information such as location, software version, sensor model fit TTN purpose and are indeed useful to monitor the health of a device, but data such as indoor/outdoor seems to be irrelevant here. They are applications oriented data and as such should be stored somewhere else or sent by the device himself.

When setting up large use case TTN allow simple communication with devices and monitoring. Data processing, storage, business intelligence will be handled by an integration or by your own solution. There the rest of metadata will be readily available or pulled from a service or database.

dakoller:
I'd vote for option 2) , because option 1) requires you to know all potential keys/values, which later can be applied.
This limits the usecases, which other people might want to do with TTN.
In my understanding the solution should allow to stick all sensor-specific metadata to the node object in TTN.

Sypheos:
We will move on with option 2.

Here is the plan:
1. Add only reserved field, they will be generic data that can be filled by TTN or the user. Most likely to be under the form 'ttn_..."
2. Customs keys will be added later but limited. Having a few keys can indeed prove very useful but providing more is not TTN job. That's why we have integration with more complex platform such as Azure, AWS and other IoT broker.

I will open issues for each development phase and reference them here.

Sypheos:
[...]

PR #588, opened 2017-05-03
Feature/nodes metadata

Sypheos:
Adding the Custom key/value map for devices. #507 

Users will be able to manage custom metadata for their device through ttnctl:

- Set: `ttnctl devices set <dev_id> --customs-keys key1:v1,key2:v2`
- See: `ttnctl devices info <dev_id`

For now their is no way to remove a key from the map you can attribute `nil` value.
- Nil a value `ttnctl devices set <dev_id> --customs-keys key1:`

The whitelist have to be manually set in redis. Config file is coming !
whitelist model: `k1:k2....`

The pull request also change several files to reduce the number lines in functions and augment maintainability.

Tests are coming...


api/handler/handler.proto
```diff
@@ -91,6 +91,7 @@ message Device {
   float latitude  = 10;
   float longitude = 11;
   int32 altitude  = 12;
+  map<string, string> custom_keys = 13;
```
johanstokking: `attributes` ?

core/handler/device/store.go
```diff
@@ -39,18 +41,22 @@ func NewRedisDeviceStore(client *redis.Client, prefix string) *RedisDeviceStore
 	for v, f := range migrate.DeviceMigrations(prefix) {
 		store.AddMigration(v, f)
 	}
+	s, _ := client.Get(prefix + ":" + redisDevicePrefix + ":whitelist").Result()
+	l := strings.Split(s, ":")
```
johanstokking: - Error handling
- Use set if it's a set

core/handler/manager_utils.go
```diff
@@ -0,0 +1,65 @@
...
+	nsUpdated := dev.GetLoRaWAN()
+	nsUpdated.FCntUp = lorawan.FCntUp
+	nsUpdated.FCntDown = lorawan.FCntDown
+	_, err := h.deviceManager.SetDevice(ctx, nsUpdated)
+	return err
+}
+
+//ctlCustomsKeys remove all the non-whitelisted customsKeys for device by creating a new key map
+func (h *handlerManager) checkCustomsKeys(in *pb.Device) {
+
```
johanstokking: Remove empty line

core/handler/manager_utils.go
```diff
@@ -0,0 +1,65 @@
...
+func (h *handlerManager) updateDevBrk(ctx context.Context, dev *device.Device, lorawan *pb_lorawan.Device) error {
+	nsUpdated := dev.GetLoRaWAN()
+	nsUpdated.FCntUp = lorawan.FCntUp
+	nsUpdated.FCntDown = lorawan.FCntDown
+	_, err := h.deviceManager.SetDevice(ctx, nsUpdated)
+	return err
+}
+
+//ctlCustomsKeys remove all the non-whitelisted customsKeys for device by creating a new key map
+func (h *handlerManager) checkCustomsKeys(in *pb.Device) {
```
johanstokking: No plural customs

core/handler/manager_utils.go
```diff
@@ -0,0 +1,65 @@
...
+//updateDevBrk Update the device in the Broker (NetworkServer)
+func (h *handlerManager) updateDevBrk(ctx context.Context, dev *device.Device, lorawan *pb_lorawan.Device) error {
+	nsUpdated := dev.GetLoRaWAN()
+	nsUpdated.FCntUp = lorawan.FCntUp
+	nsUpdated.FCntDown = lorawan.FCntDown
+	_, err := h.deviceManager.SetDevice(ctx, nsUpdated)
+	return err
+}
+
+//ctlCustomsKeys remove all the non-whitelisted customsKeys for device by creating a new key map
```
johanstokking: Comment doesn't match function name

core/handler/device/store.go
johanstokking: Whitelist is too generic imo. It should be clear that it's a filter for attribute keys that are published upstream

Sypheos: No idea on how to handle an error there... apart for fmt.Printf(err)

The whitelist is stored as a simple string. Each key is separated by ":" so simply split it into a slice will do. Using set is overkill for this.

ttnctl/cmd/devices_set.go
```diff
@@ -153,13 +153,17 @@ var devicesSetCmd = &cobra.Command{
...
 				s := strings.SplitN(v, ":", 2)
-				if dev.CustomKeys == nil {
-					dev.CustomKeys = make(map[string]string, len(in))
+				if dev.Attributes == nil {
+					dev.Attributes = make(map[string]string, len(in))
+				}
+				if len(s) == 2 {
+					dev.Attributes[s[0]] = s[1]
+				} else {
+					dev.Attributes[s[0]] = ""
```
johanstokking: Is this the way to unset an attribute? That wouldn't be very intuitive

Sypheos: It's not. It's to protect from panic if someone add a key without value, the result will be an empty value for a key in the map.

Removing a custom attributes is currently not possible. Thinking about way to do it without adding more to  flags

Sypheos:
Did I missed some ?

ttnctl/cmd/devices_set.go
johanstokking: I think it's fine to have a flag. What about `--attr-add` and `--attr-remove` ?

johanstokking:
Yes, I still see `CustomKeys` and further on `customAttr` or something

core/types/uplink_message.go
```diff
@@ -15,5 +15,5 @@ type UplinkMessage struct {
 	PayloadRaw     []byte                 `json:"payload_raw"`
 	PayloadFields  map[string]interface{} `json:"payload_fields,omitempty"`
 	Metadata       Metadata               `json:"metadata,omitempty"`
-	CustomKeys     map[string]string      `json:"custom_keys"`
+	Attributes     map[string]string      `json:"attributes"`
```
johanstokking: + `omitempty`

Procfile
```diff
@@ -2,4 +2,4 @@ rtr: ttn router --config ./.env/router/dev.yml --health-port 10901
 dsc: ttn discovery --config ./.env/discovery/dev.yml --health-port 10900
 ns:  ttn networkserver --config ./.env/networkserver/dev.yml --health-port 10903
 brk: ttn broker --config ./.env/broker/dev.yml --health-port 10902
-hdl: ttn handler --config ./.env/handler/dev.yml --health-port 10904
+//hdl: ttn handler --config ./.env/handler/dev.yml --health-port 10904
```
johanstokking: Remove comment

ttnctl/cmd/devices_set.go
Sypheos: Can work with that. 

Then the attributes in ` - - attr-remove ` will made empty. 
So every attributes who arrive with empty values will be deleted. 
But it also mean that with `--attr-add` if there is no value it will delete the attribute.

Sypheos: Can work with that. 

Then the attributes in `--attr-remove ` will made empty. 
So every attributes who arrive with empty values will be deleted. 
But it also mean that with `--attr-add` if there is no value it will delete the attribute.

core/handler/manager_utils.go
```diff
@@ -0,0 +1,73 @@
...
+func (h *handlerManager) updateDevBrk(ctx context.Context, dev *device.Device, lorawan *pb_lorawan.Device) error {
+	nsUpdated := dev.GetLoRaWAN()
+	nsUpdated.FCntUp = lorawan.FCntUp
+	nsUpdated.FCntDown = lorawan.FCntDown
+	_, err := h.handler.ttnDeviceManager.SetDevice(ctx, nsUpdated)
+	return err
+}
+
+//ctlCustomsKeys take all the whitelisted Attribute plus a maximum of customs one
+func (h *handlerManager) ctlCustomsKeys(in *pb.Device) {
```
johanstokking: Rename

ttnctl/cmd/devices_set.go
```diff
@@ -204,5 +213,6 @@ func init() {
 	devicesSetCmd.Flags().Int32("altitude", 0, "Set altitude")
 
 	devicesSetCmd.Flags().String("description", "", "Set Description")
-	devicesSetCmd.Flags().StringArray("custom-attr", []string{}, "Set a custom attribute")
+	devicesSetCmd.Flags().StringArray("attr-add", []string{}, "Add an additional attribute (ex:--attr-add <name>:<value>)")
```
johanstokking: No examples in description

ttnctl/cmd/devices_set.go
```diff
@@ -204,5 +213,6 @@ func init() {
...
+	devicesSetCmd.Flags().StringArray("attr-rm", []string{}, "Remove an additional attribute")
```
johanstokking: Idem

ttnctl/cmd/devices_set.go
johanstokking: What is an _additional_ attribute other than just attribute? I don't think that ID is an attribute of a device; I think that all custom things are attributes

Sypheos: If we simply call them attributes it's confusing
I mean this attributes are like the far west. I can of course be something pretty logical like *perOfBattery* but I can be *doesTheSunTurnAroundTheEarth=cookie*.
There is no control over it and people can put whatever stuff they want and it won't always be a *device attribute*. Maybe use this data to have a teaParty on the sun, why not.
I'll rename it to *key* or *meta*. 

P.S: It just occurred to me that *attribute* might not be the name we want to give to this.

johanstokking: I think attribute is the right naming. Indeed, people can put random stuff, but that's up to the people.

We may want to restrict the key to match our generic regex for ids? That way, we can safely use it in the future

Sypheos: Like camelCase ? or restrict to a combination of some words ?

core/handler/device/store.go
```diff
@@ -23,11 +24,13 @@ type Store interface {
 	DownlinkQueue(appID, devID string) (DownlinkQueue, error)
 	Set(new *Device, properties ...string) (err error)
 	Delete(appID, devID string) error
+	AttrWhitelist() map[string]bool
 }
 
 const defaultRedisPrefix = "handler"
 const redisDevicePrefix = "device"
 const redisDownlinkQueuePrefix = "downlink"
+const redisUpStreamMetadataKeyWhitelist = ":attr-whitelist"
```
htdvisser: 1) This is just configuration, and you load it on start. Why do we need to store this in Redis, instead of using the handler's config?
2) `s/redisUpStreamMetadataKey/redisDeviceAttribute/g`
3) lose the colon, that belongs in the formatting, not here

core/handler/manager_utils.go
```diff
@@ -0,0 +1,74 @@
...
+	pb_lorawan "github.com/TheThingsNetwork/ttn/api/protocol/lorawan"
+	"github.com/TheThingsNetwork/ttn/core/handler/device"
+	"github.com/TheThingsNetwork/ttn/core/types"
+	"github.com/TheThingsNetwork/ttn/utils/errors"
+)
+
+const maxAttr uint8 = 5
+
+//eventSelect select the appropriate event for device is updated/created and check if the event is possible
+func (h *handlerManager) eventSelect(ctx context.Context, dev *device.Device, lorawan *pb_lorawan.Device, appId string) (evt types.EventType, err error) {
```
htdvisser: Unnecessary refactoring. This function is only used once. Why take it out of the normal flow?

core/handler/manager_utils.go
```diff
@@ -0,0 +1,74 @@
...
+			if existingDevice.AppEUI == *lorawan.AppEui && existingDevice.DevEUI == *lorawan.DevEui {
+				return "", errors.NewErrAlreadyExists("Device with AppEUI and DevEUI")
+			}
+		}
+	}
+	return evt, nil
+}
+
+//updateDevBrk Update the device in the Broker (NetworkServer)
+func (h *handlerManager) updateDevBrk(ctx context.Context, dev *device.Device, lorawan *pb_lorawan.Device) error {
```
htdvisser: Unnecessary refactoring. This function is only used once. Why take it out of the normal flow?

ttnctl/cmd/devices_set.go
```diff
@@ -190,4 +213,6 @@ func init() {
 	devicesSetCmd.Flags().Int32("altitude", 0, "Set altitude")
 
 	devicesSetCmd.Flags().String("description", "", "Set Description")
+	devicesSetCmd.Flags().StringArray("attr-add", []string{}, "Add an attribute")
+	devicesSetCmd.Flags().StringArray("attr-rm", []string{}, "Remove an attribute")
```
htdvisser: Use the full "remove" instead of "rm"

johanstokking:
@htdvisser yep, I was too fast, also revoked the approval later. This certainly doesn't meet quality standards yet, thanks for reviewing

core/handler/device/store.go
Sypheos: The first idea was to fetch the whitelist every time attributes are added. Then we can change which the whitelist without having to restart the handler and change a config file.
Wanted to confirm it was working like this first.

Though I am planing to add the list to the config file.

johanstokking: I think the whitelist is not very volatile. Also it shouldn't matter what's in the whitelist, except when it comes to publishing it. It's good to reconsider the name whitelist as well; it's more a filter of attributes that are sent upstream

Sypheos: then `attr-filter` / `atttribute-filter` ?

johanstokking: 1. Use the generic ID validator to check attribute keys
2. Restrict the length of attribute values to 200 bytes. Bit arbitrary but there should be a limit
3. There should be a list of builtin attributes, starting with `ttn-`, in configuration, whose set values should always be published upstream (e.g. `ttn-brand`, `ttn-model`). I think this is currently the "whitelist", but I'd call them "standard attributes", but open to suggestions
4. Other attributes should be limited by a fixed number, which is also set in the configuration. Which attributes are being published can be random or alphabetically sorted. The latter gives a consistent result

Sypheos: 1. Is there is one already or built one ?
2. 128 or 256 ?
3. That's the white list, name can be anything as long as it state that it's always allow this attributes. `common attributes` / `ttn attributes`
4. The limitation is implemented, have to take it from the config. Sort the pushed attributes, you mean sort the attributes received by the client and then push them to the db ?

johanstokking: 1. [Yes](https://github.com/TheThingsNetwork/ttn/blob/master/api/validation.go)
2. Too geeky; there's no reason for in this case, nothing Redis optimises here for example
3. Yep. For me, "standard" is "builtin" as it clearly indicates that it's standard or builtin. Common can be anything people commonly use, and everything here is already "ttn" so that may confuse. From an outside perspective, all these attributes are "ttn attributes"
4. Get the attributes of the device, send the standard ones anyway, sort the other ones alphabetically and select the top n from them.

Sypheos: 1. ok
2. Then 200 ?
3. "builtin" rings good
4. For the sorting, should there be a mechanism that limit the number of key transmitted? Too many values to sort will make the processors spikes. I think gRPC have a maximum size for the packet no ? or brutally trunc if they are too many.

johanstokking: 2. Yes
4. Maybe you can already limit the number of custom attributes when getting them out of the data store? This is probably also the same on repeats. What I'd like to avoid is that if people have more custom attributes than we allow to publish upstream, that on each message the attributes are different (e.g. if we use a `map[string]string` and break the range after n, the result is random)

Sypheos: 4. I see. Then prioritize the one already on the store and after take the first n till maximum allowed ?

johanstokking: Yes

core/handler/manager_utils.go
Sypheos: Split the function for better readability and tests. Limiting lines in one functions.

Sypheos:
Just added tests. Sorry should have made them earlier, always planned to do them.

.env/handler/dev.yml
```diff
@@ -8,5 +8,6 @@ key-dir: "./.env/handler/"
 handler:
   amqp-address: localhost:5672
   mqtt-address: localhost:1883
+  builtin-attr: ttn-Latitude
```
johanstokking: Keep lowercase

johanstokking: Keep lowercase

core/handler/device/converter_test.go
```diff
@@ -0,0 +1,72 @@
+// Copyright © 2017 The Things Industries B.V.
```
johanstokking: Wrong copyright. Please check all

core/handler/device/store.go
```diff
@@ -23,6 +24,8 @@ type Store interface {
 	DownlinkQueue(appID, devID string) (DownlinkQueue, error)
 	Set(new *Device, properties ...string) (err error)
 	Delete(appID, devID string) error
+	SetBuiltinAttr(string)
+	GetAttrWhitelist() map[string]bool
```
johanstokking: What's the difference between builtin and whitelist?

core/handler/device/store.go
```diff
@@ -49,8 +52,9 @@ func NewRedisDeviceStore(client *redis.Client, prefix string) *RedisDeviceStore
 // RedisDeviceStore stores Devices in Redis.
 // - Devices are stored as a Hash
 type RedisDeviceStore struct {
-	store  *storage.RedisMapStore
-	queues *storage.RedisQueueStore
+	store         *storage.RedisMapStore
+	queues        *storage.RedisQueueStore
+	attrWhitelist map[string]bool
```
johanstokking: Builtin

johanstokking: Please checks diffs on inconsistencies; this should be builtin

core/handler/manager_utils.go
```diff
@@ -0,0 +1,80 @@
...
+	nsUpdated := dev.GetLoRaWAN()
+	nsUpdated.FCntUp = lorawan.FCntUp
+	nsUpdated.FCntDown = lorawan.FCntDown
+	_, err := h.handler.ttnDeviceManager.SetDevice(ctx, nsUpdated)
+	return err
+}
+
+//attrControl take all the whitelisted Attribute plus a maximum of customs one
+func (h *handlerManager) attrControl(in *pb.Device) {
```
johanstokking: Whitelist? What does it _control_?

Also is it logical that it overwrites the attributes here?

Issue #590, opened 2017-05-04
Decode function, payload_fields cannot be converted to json


Sypheos:
This is a **bug report** for **the backend**.

When using customs decode functions to parse the payload_raw and obtain payload_fields the obtained `map[string]interface{}` might not be able to be converted to json format.
Such as some results from floating pointer arithmetic can be impossible to parse for the golang json parser.

PR #591, opened 2017-05-04
Checking if payload fields can be marshalled or not

Sypheos:
#590


htdvisser:
Again a PR without tests. Can't accept that. I changed a few things and added tests.

Issue #420, opened 2017-01-17
Refactor Router Gateways/Schedule


htdvisser:
- [ ] Should be able to clean up when gtw disconnects
- [ ] Schedule should not be responsible for handling multiple subscribers
- [ ] Schedule should respect RTT
- [ ] Schedule should use less goroutines

htdvisser:
Won't do this for v2

PR #610, opened 2017-05-22
Refactor downlink logic in Router

htdvisser:
This PR addresses a number of issues:

- #420: work with gateway's RTT.
- #420: use less goroutines (now goroutines are created per schedule instead of per scheduled message).
- #420: schedule is no longer responsible for handling multiple subscriptions.
- #445: duty cycle and utilization have been moved to the frequency plan.


core/router/gateway/downlink.go
```diff
@@ -0,0 +1,212 @@
+package gateway
```
johanstokking: Copyright

htdvisser:
Will reopen later

Issue #572, opened 2017-04-24
Send component status messages to the NOC


rvolosatovs:
This is a **feature request** for **the backend**.

Router, handler and broker status messages should be sent to the NOC.

htdvisser:
Let's also add the networkserver here

htdvisser:
This will be streams of

- [`router.Status`](https://github.com/TheThingsNetwork/ttn/blob/develop/api/router/router.proto#L89-L101)
- [`broker.Status`](https://github.com/TheThingsNetwork/ttn/blob/develop/api/broker/broker.proto#L168-L182)
- [`networkserver.Status`](https://github.com/TheThingsNetwork/ttn/blob/develop/api/networkserver/networkserver.proto#L49-L58)
- [`handler.Status`](https://github.com/TheThingsNetwork/ttn/blob/develop/api/handler/handler.proto#L37-L44)

htdvisser:
@gomezjdaniel: please discuss with @rvolosatovs if component statuses can already be sent to the NOC. If ok, please make sure the routing components send their status to the NOC every X seconds. 

For the Handler, it's probably just a `time.Tick` and `monitorStream.Send(h.GetStatus())`, but for the Router and Networkserver you might have to add some initialization (just copy-paste)

htdvisser:
@jankramer: Does vx support the component status RPCs?

jankramer:
@htdvisser no, the plan is to get these from the NOC instead.

rvolosatovs:
The component statuses can be sent to the NOC once @htdvisser approves TheThingsIndustries/noc#197 and I deploy
They will not get stored yet though (that's TheThingsIndustries/noc#190)

PR #623, opened 2017-06-08
Monitor handler, broker, router and NS statuses

gomezjdaniel:
The PR gives the ability to Router/Handler/NetworkServer/Broker to send status messages to different monitors in intervals of time.

A default interval value of 6 seconds is set, https://github.com/TheThingsNetwork/ttn/pull/623/files#diff-ff7686b39bf90dc2520886fb874371a4R139

PR references to https://github.com/TheThingsNetwork/ttn/issues/572


cmd/root.go
```diff
@@ -136,6 +136,8 @@ func init() {
 		"ttn-account-v2": "https://account.thethingsnetwork.org",
 	})
 
+	viper.SetDefault("monitor-interval", "6")
```
htdvisser: Add this under the `PersistentFlags` of the root cmd

core/broker/broker.go
```diff
@@ -135,6 +136,10 @@ func (b *broker) Init(c *component.Component) error {
 	b.Component.SetStatus(component.StatusHealthy)
 	if b.Component.Monitor != nil {
 		b.monitorStream = b.Component.Monitor.NewBrokerStreams(b.Identity.Id, b.AccessToken)
+
+		if len(viper.GetStringMapString("monitor-servers")) > 0 {
```
htdvisser: I'm not a big fan of reading config "by string" somewhere deep in the code... Same in the other files

vendor/vendor.json
```diff
@@ -718,7 +718,7 @@
 			"revisionTime": "2017-05-10T02:47:26Z"
 		},
 		{
-			"checksumSHA1": "oBgjrQKzT1cPwaRQy4QSgRCqSeo=",
```
htdvisser: Not sure what happened here.. Did you change something or is this something strange in Windows vs Mac/Linux?

core/broker/broker.go
gomezjdaniel: What do you suggest then to avoid running this goroutine (with an infinite loop inside) when there is no monitor-servers specified?

Here there is a similar case reading from config elsewhere that's not cmd: [core/component/status.go](https://github.com/TheThingsNetwork/ttn/blob/4db1a6ff72d07f03f5e194d28ab44439bdf1648a/core/component/status.go#L90)

htdvisser: Let's leave it like this until we refactor the configuration system (do you agree, @johanstokking?)

johanstokking: Part of the cleanup refactor would be depending on config system (i.e. viper, cobra, yml rewriting) only in `cmd` or `utils` package and initialize the components in a clean way.

I'd rather not merge known technical debt, so better fix this. It's an easy fix.

PR #628, opened 2017-06-17
Types package randomizers

rvolosatovs:
Lack of these randomizers prevent using gogoproto `populate` option to auto-generate randomizers for types in NOC. This implementation does not add any additional dependencies to the `core/types` package, (test code is in separate `types_test` package) as was determined to be a requirement in previous discussion with @htdvisser.
`Read` function is not used, as interface generated by `gogoproto` does not provide it and `Int63()` is used under the hood anyway. Pointers are returned to match `gogoproto` expectations.

### DIFF
--- a/core/types/random.go
+++ b/core/types/random.go
@@ -29,7 +29,8 @@ func randRead(r Rand, b []byte) (n int, err error) {
 }
 
 // NewPopulatedDevAddr returns random DevAddr
-func NewPopulatedDevAddr(r Rand) (devAddr DevAddr) {
+func NewPopulatedDevAddr(r Rand) (devAddr *DevAddr) {
+	devAddr = &DevAddr{}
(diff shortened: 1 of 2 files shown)


core/types/random.go
```diff
@@ -0,0 +1,56 @@
...
+		pos--
+	}
+	return
+}
+
+// NewPopulatedDevAddr returns random DevAddr
+func NewPopulatedDevAddr(r Rand) (devAddr *DevAddr) {
+	devAddr = &DevAddr{}
+	if _, err := randRead(r, devAddr[:]); err != nil {
+		panic(fmt.Errorf("types.NewPopulatedAppEUI: %s", err))
```
htdvisser: Panic should say `NewPopulatedDevAddr`

htdvisser:
Squash-merged in 3715af852af7c424fc883c750d5bf5034a372f55

PR #635, opened 2017-06-22
Add location source to metadata

htdvisser:
This PR adds the source of the gateway (and device) location to the protobuf and MQTT messages

⚠️ API change (cc: @romeovs @rvolosatovs @dotpy3) in gateway protobuf (binary compatible):
- `gps` was **renamed** to `location` and `GpsMetadata` was **renamed** to `LocationMetadata`
- `source` was **added** to `LocationMetadata` 

⚠️ API change in MQTT json:
- `location_source` was **added** to `LocationMetadata`

TODO:

- Insert location source in packet forwarder - @dotpy3 
- Insert location source in bridge - @htdvisser
- Process location source in NOC - @rvolosatovs 
- Process location source in Console - @romeovs 
- Process location source in TTNMapper - @jpmeijers


api/gateway/gateway.proto
```diff
@@ -9,13 +9,23 @@ package gateway;
...
+
+  LocationSource source = 5;
+
+  enum LocationSource {
+    UNKNOWN  = 0;
+    GPS      = 1;
+    CONFIG   = 2;
+    REGISTRY = 3;
+    ESTIMATE = 4;
+  }
```
dotpy3: Is the meaning of every value documented?

api/gateway/validation.go
```diff
@@ -23,14 +23,30 @@ func (m *Status) Validate() error {
...
 	return nil
 }
 
-func (m GPSMetadata) IsZero() bool {
-	return m.Latitude == 0 && m.Longitude == 0
+const delta = 0.01
+
+// IsZero returns whether the location is close enough to zero (and should be nil)
+func (m LocationMetadata) IsZero() bool {
+	return (m.Latitude > -delta && m.Latitude < delta) && (m.Longitude > -delta && m.Longitude < delta) && m.Altitude == 0
```
jpmeijers: Maybe we should not check the altitude value here. If the location is on Null island, it is incorrect, even if we have a valid altitude.

api/gateway/gateway.proto
```diff
@@ -9,13 +9,28 @@ package gateway;
...
   int32   altitude   = 4;
+
+  LocationSource source = 5;
+
+  enum LocationSource {
+    // The source of the location is not known or not set
+    UNKNOWN  = 0;
+    // The location is determined by GPS
+    GPS      = 1;
+    // The location is fixed by configuration
```
jpmeijers: By configuration we mean "fake gps"?

api/gateway/gateway.proto
```diff
@@ -9,13 +9,28 @@ package gateway;
...
+    CONFIG   = 2;
+    // The location is set in and updated from a registry
+    REGISTRY = 3;
+    // The location is estimated
```
jpmeijers: Sounds like a good feature to have. But shouldn't we specify how it was estimated? From IP, from coverage, etc.

api/gateway/validation.go
```diff
@@ -23,14 +23,30 @@ func (m *Status) Validate() error {
...
+
 // Validate implements the api.Validator interface
-func (m *GPSMetadata) Validate() error {
-	if m == nil || m.IsZero() {
-		return errors.NewErrInvalidArgument("GPSMetadata", "can not be empty")
+func (m *LocationMetadata) Validate() error {
+	if m.IsZero() {
+		return ErrLocationZero
+	}
+	if m.Latitude > 90 || m.Latitude < -90 {
```
jpmeijers: I would even consider writing it like this:
`if m.Latitude > (90 - delta) || m.Latitude < (-90 + delta) {`

Because it is really very unlikely for a gateway to be located exactly on the pole, and is most likely a miss-configuration (that I've see a few times).

api/gateway/gateway.proto
htdvisser: for example

api/gateway/validation.go
htdvisser: Agreed

api/gateway/validation.go
htdvisser: I agree

api/gateway/gateway.proto
htdvisser: In that case I think we should leave it out for now, and add specific Enum values for different estimation techniques when we implement them. Unless we can come up with a decent list of techniques we expect to use.

jpmeijers: Maybe a good idea to leave it out for now then. Or add one as example: "IP_GEOLOCATION".

api/gateway/validation.go
```diff
@@ -23,14 +23,30 @@ func (m *Status) Validate() error {
...
-	if m == nil || m.IsZero() {
-		return errors.NewErrInvalidArgument("GPSMetadata", "can not be empty")
+func (m LocationMetadata) Validate() error {
+	if m.IsZero() {
+		return ErrLocationZero
+	}
+	if m.Latitude >= 90-delta || m.Latitude <= -90+delta {
+		return ErrInvalidLatitude
+	}
+	if m.Longitude >= 180-delta || m.Longitude <= -180+delta {
```
jpmeijers: I don't think we should add the delta for the longitude. The chance is much higher for a gateway standing on or close to the -180/180 (international date) line. For -90/+90 it is a point (the poles), for which the probability is extremely low.

htdvisser:
@jpmeijers @johanstokking @dotpy3 approved?

PR #647, opened 2017-07-05
Add save option to `authorize`

dotpy3:
For programmatic deployments of The Things Network, we need to be able to generate tokens and to get them. Currently, when tokens are generated, they are only shown on the command output - which means getting the token requires parsing the output. It's not sustainable, since the CLI output might change over time.
This PR introduces a new option to the `authorize` commands, `--save`, which allows the user to point to a file where to save the token. This way, when programmatically deploying The Things Network, it is possible to retrieve the token safely. If this is not satisfying, we could imagine an option to hide the logs from the output.


htdvisser:
Isn't it easier to just remove the unneeded log lines and just print the token directly? These cmds are for expert users anyway.

dotpy3:
PR closed because the needs are already fulfilled by `--no-cli-logs`.

PR #650, opened 2017-07-10
Clean up monitor package

htdvisser:


htdvisser:
@rvolosatovs

api/monitor/monitorclient/client.go
```diff
@@ -0,0 +1,217 @@
...
+	c.setup()
+}
+
+func (c *componentClient) Send(msg interface{}) {
+	switch msg := msg.(type) {
+	case *gateway.Status, *router.Status, *broker.Status, *networkserver.Status, *handler.Status:
+		if len(c.status) == 0 {
+			return
+		}
+		for _, cli := range c.status {
```
rvolosatovs: can we make a convention to use `cl` for client? `cli` is really ambiguous

api/monitor/monitorclient/gateway.go
```diff
@@ -0,0 +1,49 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// GatewayClient creates a new client for Gateway monitoring
+func (m *MonitorClient) GatewayClient(ctx context.Context, opts ...grpc.CallOption) Stream {
+	c := new(componentClient)
+	c.log = m.log.WithField("Component", "Gateway")
+	if md, ok := metadata.FromOutgoingContext(ctx); ok {
+		if id, ok := md["id"]; ok && len(id) > 0 {
```
rvolosatovs: `ttnctx.IDFromMetadata`

api/monitor/monitorclient/handler.go
```diff
@@ -0,0 +1,49 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// HandlerClient creates a new client for Handler monitoring
+func (m *MonitorClient) HandlerClient(ctx context.Context, opts ...grpc.CallOption) Stream {
+	c := new(componentClient)
+	c.log = m.log.WithField("Component", "Handler")
+	if md, ok := metadata.FromOutgoingContext(ctx); ok {
+		if id, ok := md["id"]; ok && len(id) > 0 {
```
rvolosatovs: `ttnctx.IDFromMetadata`

api/monitor/monitorclient/networkserver.go
```diff
@@ -0,0 +1,37 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// NetworkServerClient creates a new client for NetworkServer monitoring
+func (m *MonitorClient) NetworkServerClient(ctx context.Context, opts ...grpc.CallOption) Stream {
+	c := new(componentClient)
+	c.log = m.log.WithField("Component", "NetworkServer")
+	if md, ok := metadata.FromOutgoingContext(ctx); ok {
+		if id, ok := md["id"]; ok && len(id) > 0 {
```
rvolosatovs: `ttnctx.IDFromMetadata`

api/monitor/monitorclient/router.go
```diff
@@ -0,0 +1,37 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// RouterClient creates a new client for Router monitoring
+func (m *MonitorClient) RouterClient(ctx context.Context, opts ...grpc.CallOption) Stream {
+	c := new(componentClient)
+	c.log = m.log.WithField("Component", "Router")
+	if md, ok := metadata.FromOutgoingContext(ctx); ok {
+		if id, ok := md["id"]; ok && len(id) > 0 {
```
rvolosatovs: `ttnctx.IDFromMetadata`

core/router/gateway/gateway.go
```diff
@@ -51,18 +50,17 @@ func (g *Gateway) SetAuth(token string, authenticated bool) {
...
 	}
 	g.token = token
-	if g.Monitor != nil {
-		g.Ctx.Debug("Starting Gateway Monitor Stream")
-		g.MonitorStream = g.Monitor.NewGatewayStreams(g.ID, g.token)
-	}
+	g.MonitorStream.Reset()
+}
+
+func (g *Gateway) GetToken() string {
```
rvolosatovs: `Token()`

core/router/gateway/gateway.go
```diff
@@ -51,18 +50,17 @@ func (g *Gateway) SetAuth(token string, authenticated bool) {
...
+	g.mu.RLock()
+	defer g.mu.RUnlock()
```
rvolosatovs: This function is short enough for `defer` to produce significant overhead.
https://lk4d4.darth.io/posts/defer/

```
g.mu.RLock()
token := g.token
g.mu.RUnlock()
return token
```

api/monitor/monitorclient/router.go
```diff
@@ -0,0 +1,37 @@
...
+			c.log = c.log.WithField("ID", id[0])
+		}
+	}
+	c.setup = func() {
+		ctx, c.cancel = context.WithCancel(ctx)
+		for name, cli := range m.clients {
+			name, cli := name, cli // shadow vars
+
+			status := sendbuffer.New(m.bufferSize, func() (grpc.ClientStream, error) {
```
rvolosatovs: names like this make it hard to undestand, what's actually going on IMO. why not call it `stream`,  `buf`/`buffer`/`b`, something in that direction? I have no clue what `sendbuffer` does, so I leave that up to you.

api/monitor/monitorclient/networkserver.go
```diff
@@ -0,0 +1,37 @@
...
+			c.log = c.log.WithField("ID", id[0])
+		}
+	}
+	c.setup = func() {
+		ctx, c.cancel = context.WithCancel(ctx)
+		for name, cli := range m.clients {
+			name, cli := name, cli // shadow vars
+
+			status := sendbuffer.New(m.bufferSize, func() (grpc.ClientStream, error) {
```
rvolosatovs: names like this make it hard to undestand, what's actually going on IMO. why not call it `stream`,  `buf`/`buffer`/`b`, something in that direction? I have no clue what `sendbuffer` does, so I leave that up to you.

api/monitor/monitorclient/handler.go
```diff
@@ -0,0 +1,49 @@
...
+			c.log = c.log.WithField("ID", id[0])
+		}
+	}
+	c.setup = func() {
+		ctx, c.cancel = context.WithCancel(ctx)
+		for name, cli := range m.clients {
+			name, cli := name, cli // shadow vars
+
+			uplink := sendbuffer.New(m.bufferSize, func() (grpc.ClientStream, error) {
```
rvolosatovs: names like this make it hard to undestand, what's actually going on IMO. why not call it `stream`,  `buf`/`buffer`/`b`, something in that direction? I have no clue what `sendbuffer` does, so I leave that up to you.

api/monitor/monitorclient/handler.go
```diff
@@ -0,0 +1,49 @@
...
+	c.log = m.log.WithField("Component", "Handler")
+	if md, ok := metadata.FromOutgoingContext(ctx); ok {
+		if id, ok := md["id"]; ok && len(id) > 0 {
+			c.log = c.log.WithField("ID", id[0])
+		}
+	}
+	c.setup = func() {
+		ctx, c.cancel = context.WithCancel(ctx)
+		for name, cli := range m.clients {
+			name, cli := name, cli // shadow vars
```
rvolosatovs: not required
those are evaluated every iteration in all the statements.
unless you do `go func() {/*use cli and name*/}()` it should not cause problems

api/monitor/monitorclient/client.go
```diff
@@ -0,0 +1,217 @@
...
+			cli.SendMsg(msg)
+		}
+		c.log.Debug("Forwarded downlink to monitor")
+	default:
+		c.log.Warnf("Unknown message type: %T", msg)
+	}
+}
+
+func (c *componentClient) run(monitor, stream string, buf *sendbuffer.Stream) {
+	if err := buf.Run(); err != nil && err != context.Canceled {
```
rvolosatovs: I have no clue what `sendbuffer` package does(that's out of scope of my review, I suppose), comment says `Run the stream`, which is not too helpful. Probably OK.

api/monitor/monitorclient/client.go
```diff
@@ -0,0 +1,217 @@
...
+// MonitorClient is an overlay on top of the gRPC client
+type MonitorClient struct {
+	log          log.Interface
+	bufferSize   int
+	clients      map[string]monitor.MonitorClient
+	statusTicker *time.Ticker
+}
+
+// TickStatus calls f periodically if the StatusTicker has been set
+func (m *MonitorClient) TickStatus(f func()) {
```
rvolosatovs: 1. Why make it so specific to status?
2. Does it really belong here anyway? Why not make this functionality more generic and put it in `core`?
3. Assuming Ticker is set, it is never stopped, so this function, if ran in goroutine(how it's supposed to be, I assume) never returns

core/broker/broker.go
```diff
@@ -134,10 +135,8 @@ func (b *broker) Init(c *component.Component) error {
 	b.checkPrefixAnnouncements()
 	b.Component.SetStatus(component.StatusHealthy)
 	if b.Component.Monitor != nil {
-		b.monitorStream = b.Component.Monitor.NewBrokerStreams(b.Identity.Id, b.AccessToken)
-		go b.Component.Monitor.TickStatus(func() {
-			b.monitorStream.Send(b.GetStatus())
-		})
+		b.monitorStream = b.Component.Monitor.BrokerClient(b.Context, grpc.PerRPCCredentials(auth.WithStaticToken(b.AccessToken)))
+		go b.Component.Monitor.TickStatus(func() { b.monitorStream.Send(b.GetStatus()) })
```
rvolosatovs: this goroutine never returns. Which may cause trouble. What if the component is down/disconnected? Do we need to query it's status all the time and send it?

core/handler/handler.go
```diff
@@ -170,10 +171,8 @@ func (h *handler) Init(c *component.Component) error {
 
 	h.Component.SetStatus(component.StatusHealthy)
 	if h.Component.Monitor != nil {
-		h.monitorStream = h.Component.Monitor.NewHandlerStreams(h.Identity.Id, h.AccessToken)
-		go h.Component.Monitor.TickStatus(func() {
-			h.monitorStream.Send(h.GetStatus())
-		})
+		h.monitorStream = h.Component.Monitor.BrokerClient(h.Context, grpc.PerRPCCredentials(auth.WithStaticToken(h.AccessToken)))
+		go h.Component.Monitor.TickStatus(func() { h.monitorStream.Send(h.GetStatus()) })
```
rvolosatovs: this goroutine never returns. Which may cause trouble. What if the component is down/disconnected? Do we need to query it's status all the time and send it?

core/networkserver/networkserver.go
```diff
@@ -87,10 +89,8 @@ func (n *networkServer) Init(c *component.Component) error {
 	}
 	n.Component.SetStatus(component.StatusHealthy)
 	if n.Component.Monitor != nil {
-		n.monitorStream = n.Component.Monitor.NewNetworkServerStreams(n.Identity.Id, n.AccessToken)
-		go n.Component.Monitor.TickStatus(func() {
-			n.monitorStream.Send(n.GetStatus())
-		})
+		n.monitorStream = n.Component.Monitor.NetworkServerClient(n.Context, grpc.PerRPCCredentials(auth.WithStaticToken(n.AccessToken)))
+		go n.Component.Monitor.TickStatus(func() { n.monitorStream.Send(n.GetStatus()) })
```
rvolosatovs: this goroutine never returns. Which may cause trouble. What if the component is down/disconnected? Do we need to query it's status all the time and send it?

core/router/router.go
```diff
@@ -93,10 +94,8 @@ func (r *router) Init(c *component.Component) error {
 	}()
 	r.Component.SetStatus(component.StatusHealthy)
 	if r.Component.Monitor != nil {
-		r.monitorStream = r.Component.Monitor.NewRouterStreams(r.Identity.Id, r.AccessToken)
-		go r.Component.Monitor.TickStatus(func() {
-			r.monitorStream.Send(r.GetStatus())
-		})
+		r.monitorStream = r.Component.Monitor.RouterClient(r.Context, grpc.PerRPCCredentials(auth.WithStaticToken(r.AccessToken)))
+		go r.Component.Monitor.TickStatus(func() { r.monitorStream.Send(r.GetStatus()) })
```
rvolosatovs: this goroutine never returns. Which may cause trouble. What if the component is down/disconnected? Do we need to query it's status all the time and send it?

api/monitor/monitorclient/client.go
htdvisser: This func is for components that send their status on the stream every X seconds (for example: router, broker, networkserver, handler). I agree that it might be better to put this in the core

api/monitor/monitorclient/client.go
htdvisser: `cli` is the convention in `ttn`, and now is not the time to change those conventions

api/monitor/monitorclient/client.go
htdvisser: https://github.com/TheThingsNetwork/go-utils/pull/33
Requested a review from you ;)

api/monitor/monitorclient/handler.go
htdvisser: Thanks! 

Ref: https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

core/router/gateway/gateway.go
htdvisser: Cool

api/monitor/monitorclient/client.go
```diff
@@ -191,8 +196,26 @@ func (c *componentClient) Send(msg interface{}) {
...
+
 func (c *componentClient) run(monitor, stream string, buf *streambuffer.Stream) {
-	if err := buf.Run(); err != nil && err != context.Canceled {
-		c.log.WithField("Monitor", monitor).WithError(err).Warnf("%s stream failed", buf)
+	var t *time.Timer
+	var streamErrors int32
+	for {
+		t = time.AfterFunc(stableTimeout, func() { atomic.SwapInt32(&streamErrors, 0) })
+		err := buf.Run()
+		t.Stop()
```
rvolosatovs: https://godoc.org/time#Timer.Stop
this is potential leak I assume, you should explicitly drain the channel

htdvisser: That's only for timers created with `NewTimer`, not `AfterFunc`

PR #651, opened 2017-07-10
Add Elasticsearch options

dotpy3:
This PR adds new flags related to Elasticsearch logging:

+ `--elasticsearch-prefix` - Adds a user-defined prefix to the Elasticsearch index where documents are stored.
For the moment, they are stored in the `<component>-<date>` index (e.g. `handler-2017.07.10`). If `--elasticsearch-prefix` is set, they would be stored in `<prefix>-<component>-<date>`(e.g. with `europe` prefix, `europe-handler-2017.07.10`).
+ `--elasticsearch-username` and `--elasticsearch-password` - Allow to provide credentials if the Elasticsearch instance requires authentication.


cmd/root.go
```diff
@@ -22,10 +22,10 @@ import (
 	jsonHandler "github.com/apex/log/handlers/json"
 	levelHandler "github.com/apex/log/handlers/level"
 	multiHandler "github.com/apex/log/handlers/multi"
+	"github.com/dotpy3/go-elastic"
```
htdvisser: I don't really like this. Can't we wait for the PR to be merged? In the meantime, we can also just vendor from your fork.

cmd/root.go
```diff
@@ -72,9 +72,21 @@ var RootCmd = &cobra.Command{
 			esClient.HTTPClient = &http.Client{
 				Timeout: 5 * time.Second,
 			}
+
+			username := viper.GetString("elasticsearch-username")
```
htdvisser: Let's use `esUsername`/`esPassword`, otherwise it's too generic

cmd/root.go
```diff
@@ -72,9 +72,21 @@ var RootCmd = &cobra.Command{
...
+			password := viper.GetString("elasticsearch-password")
+			if username != "" {
+				esClient.SetAuthCredentials(username, password)
+			}
+
+			indexPrefix := cmd.Name()
```
htdvisser: then also `esPrefix`

dotpy3:
Let's wait for the `tj/go-elastic` PR status to change.

htdvisser:
Blocked by https://github.com/tj/go-elastic/pull/3

PR #671, opened 2017-08-23
Feature/amqp device events consumer

Sypheos:
This pull request add the device events handling in the generic AMQP consumer.

It is mainly aimed for the integration since they listen to any messages incoming on a AMQP queue. It will give them the possibility to react to device events.

It use the routing key in the delivery to determine if the message is an event or an uplinks.
Since they can be many different messages on one amqp queue the consumer do a few checks to be sure the routing key is valid and destined to him. Otherwise the delivery is not acknowledged so it can be send to another consumer. 

Changes in the prototypes of some functions have been made for ease and consistency.

### DIFF
--- a/amqp/routing_keys.go
+++ b/amqp/routing_keys.go
@@ -9,8 +9,8 @@ import (
 	"strings"
 )
 
-const simpleWildcard = "*"
-const wildcard = "#"
+const SimpleWildcard = "*"
+const Wildcard = "#"
 
 // DeviceKeyType represents the type of a device topic
 type DeviceKeyType string
@@ -38,11 +38,11 @@ func ParseDeviceKey(key string) (*DeviceKey, error) {
 		return nil, fmt.Errorf("Invalid key format")
 	}
 	var appID string
-	if matches[1] != simpleWildcard {
+	if matches[1] != SimpleWildcard {
 		appID = matches[1]
 	}
 	var devID string
-	if matches[3] != simpleWildcard {
+	if matches[3] != SimpleWildcard {
 		devID = matches[3]
 	}
 	keyType := DeviceKeyType(matches[4])
@@ -55,16 +55,16 @@ func ParseDeviceKey(key string) (*DeviceKey, error) {
 
 // String implements the Stringer interface
 func (t DeviceKey) String() string {
-	appID := simpleWildcard
+	appID := SimpleWildcard
 	if t.AppID != "" {
 		appID = t.AppID
 	}
-	devID := simpleWildcard
+	devID := SimpleWildcard
 	if t.DevID != "" {
 		devID = t.DevID
 	}
 	if t.Type == DeviceEvents && t.Field == "" {
-		t.Field = wildcard
+		t.Field = Wildcard
 	}
 	key := fmt.Sprintf("%s.%s.%s.%s", appID, "devices", devID, t.Type)
 	if t.Type == DeviceEvents && t.Field != "" {
@@ -96,7 +96,7 @@ func ParseApplicationKey(key string) (*ApplicationKey, error) {
 		return nil, fmt.Errorf("Invalid key format")
 	}
 	var appID string
-	if matches[1] != simpleWildcard {
+	if matches[1] != SimpleWildcard {
 		appID = matches[1]
 	}
 	keyType := ApplicationKeyType(matches[2])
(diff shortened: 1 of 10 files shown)


amqp/routing_keys.go
```diff
@@ -9,8 +9,8 @@ import (
 	"strings"
 )
 
-const simpleWildcard = "*"
-const wildcard = "#"
+const SimpleWildcard = "*"
+const Wildcard = "#"
```
johanstokking: Why expose them? This PR doesn't use them outside of this package

amqp/subscriber.go
```diff
@@ -142,3 +144,48 @@ func (s *DefaultSubscriber) subscribe(key string) (<-chan AMQP.Delivery, error)
 	}
 	return s.consume(queue)
 }
+
+// ConsumeMessages consumes all messages in a specific queue and try to handle them as uplinks or events depending on the
+// routing key.
+func (s *DefaultSubscriber) ConsumeMessages(queue string, uplinkHandler UplinkHandler, eventHandler DeviceEventHandler) error {
+	messages, err := s.consume(s.name)
```
johanstokking: Shouldn't this be `s.consume(queue)`?

amqp/subscriber.go
```diff
@@ -28,7 +30,7 @@ type Subscriber interface {
 	SubscribeDeviceUplink(appID, devID string, handler UplinkHandler) error
 	SubscribeAppUplink(appID string, handler UplinkHandler) error
 	SubscribeUplink(handler UplinkHandler) error
-	ConsumeUplink(queue string, handler UplinkHandler) error
+	ConsumeMessages(queue string, uplinkHandler UplinkHandler, eventHandler DeviceEventHandler) error
```
johanstokking: What is the rationale for consuming uplinks _and_ device events at once?
Why not route them to separate queues?

amqp/routing_keys.go
Sypheos: Other packages using the amqp package need them to build their routing keys.

amqp/subscriber.go
Sypheos: Didn't payed attention when I copy pasted the previous code from `ConsumeUplink`.

amqp/subscriber.go
Sypheos: One queue, one consumer, one binding, one process.
Of course it is possible to have two queues and managing more bindings which is probably faster but harder to manage.

amqp/routing_keys.go
johanstokking: How so, can they provide their own routing key?

amqp/subscriber.go
johanstokking: How so harder to manage? The original `ConsumeUplink()` was pretty straight forward. The difference would be the routing key. AMQP is _only_ used in performance critical scenarios. I don't want to nop-handle all device events in _most_ integrations because it's easier to implement.

Sypheos: It is not easier to implement both are straightforward.
It harder to manage since it will require more configuration and management, for example the integration controller will have to manage two queue and the process to subscribe to both.
The difference is minimal in term of complexity, but if it aimed for performance then it's more efficient to use two queues and depend on the broker.

johanstokking: OK I see about the configuration and the queue.

Can we bind twice to the same queue, for both routing keys?

Sypheos: A queue can have any number of bindings to any exchange. The sole problem is for the consumer. Their is no guarantee that an uplink message will arrive to the right consumer the first time. 
If we have on type of consumer for uplinks and one for events depending on the number of consumer the message might bounce back a lot of time before going to the right consumer. 

With AMQP queue consumer should all do the same thing and the queue will balance the load. So we either put the charge of routing uplinks and event to AMQP with two queues which is faster but add more configuration etc...Or we have one queue and it's the consumer that call different handler depending on the routing key.

johanstokking: OK yeah, I see the need for one queue. So when creating and binding the queue, in the controller, the right topic(s) are chosen. The process, indeed, needs to have a generic consumer. Integrations that are not interested in device events would pass a nil handler for events, in fact there wouldn't be events bound to the queue

Sypheos: Yes. If they want the events they pass a handler and add bind the queue to receive events.

Issue #619, opened 2017-06-06
Add ADR support for other bands than just EU


htdvisser:
We currently have ADR support for the EU region, but it shouldn't be too hard to implement this for the other (US, AU, AS, KR) regions as well. @gomezjdaniel: please work with @tftelkamp to implement this. Shouldn't be too complicated, maybe even as simple as copy-pasting the EU settings (https://github.com/TheThingsNetwork/ttn/blob/develop/core/band/band.go#L90) and adapting them for the other regions.

hoonppark:
To add ADR support for the Korean region,  I'd like to suggest the following line to be inserted at line 149 of 'band.go' file:
https://github.com/TheThingsNetwork/ttn/blob/develop/core/band/band.go#L149 :

Line # 149: frequencyPlan.ADR = &ADRConfig{MinDataRate: 0, MaxDataRate: 5, MinTXPower: 2, MaxTXPower: 14}

The above line is just a copy of the line # 90 (for EU ADR).

I want someone to verify if the above line has correct parameters for the Korean region ADR.  Please verify the parameters with the LoRaWAN regional spec summary for Korea listed below.

According to the LoRaWAN Spec 1.0.2 Page 40, Data Rates and TXPower for Korea (KR Data Rates and KR TXPower) are as follows:

(1) KR Data Rates -- (from LoRaWAN Spec 1.0.2 Page 40)
----------------------------------------------------------------
Data Rate	           Spreading Factor, Bandwidth
----------------------------------------------------------------
DR0 (   250 bps)	- SF12  BW125 KHz  
DR1 (   440 bps)	- SF11   BW125 KHz
DR2 (   980 bps)	- SF10   BW125 kHz
DR3 (  1760 bps)	- SF9   BW125 KHz
DR4 (  3125 bps)	- SF8   BW125 KHz

DR5 (  5470 bps)	- SF7   BW125 KHz
DR6 ~ 15  : RFU (Reserved for Future Use)
----------------------------------------------------------------


(2) KR TXPower -- (from LoRaWAN Spec 1.0.2 Page 40)
---------------------------------------------------------------
TX0 	  : 20 dbm if supported  (30 dbm = 100 mW)
TX1 	  : 14 dbm (25   mW)
TX2 	  : 10 dbm (10   mW)
TX3 	  :  8 dbm (  6   mW)
TX4 	  :  5 dbm (  3.2 mW)
TX5 	  :  2 dbm (  1.6 mW)
[...]

PR #674, opened 2017-08-27
Add support for ADR on US_902_928 band

mgranberry:
This pull request adds support for ADR on the US 902-928 band.  The power limits are capped at 20dBm, although the 500kHz channels appear to allow 26dBm.

Related to https://github.com/TheThingsNetwork/ttn/issues/619

### DIFF
--- a/core/band/band.go
+++ b/core/band/band.go
@@ -87,21 +87,16 @@ func Get(region string) (frequencyPlan FrequencyPlan, err error) {
 		}
 		frequencyPlan.DownlinkChannels = frequencyPlan.UplinkChannels
 		frequencyPlan.CFList = &lorawan.CFList{867100000, 867300000, 867500000, 867700000, 867900000}
-		frequencyPlan.ADR = &ADRConfig{MinDataRate: 0, MaxDataRate: 5, MinTXPower: 2, MaxTXPower: 14}
+		frequencyPlan.ADR = &ADRConfig{MinDataRate: 0, MaxDataRate: 5, MinTXPower: 2, MaxTXPower: 14, StepTXPower: 3}
 	case pb_lorawan.FrequencyPlan_US_902_928.String():
 		frequencyPlan.Band, err = lora.GetConfig(lora.US_902_928, false, lorawan.DwellTime400ms)
-		channelGroups := [](int){1} // Enable 903.9-905.3/200 kHz, 904.6 mHz channels
+		fsb := 1 // Enable 903.9-905.3/200 kHz, 904.6 mHz channels
 		for channel := 0; channel < 72; channel++ {
-			for _, channelGroup := range channelGroups {
-				if (channel < channelGroup*8 || channel >= (channelGroup+1)*8) && channel != channelGroup+64 {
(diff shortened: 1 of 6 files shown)


core/networkserver/adr.go
```diff
@@ -168,36 +169,102 @@ func (n *networkServer) handleDownlinkADR(message *pb_broker.DownlinkMessage, de
...
+			sort.Ints(channels)
+
+			chMaskCntl := -1
+			for _, c := range channels {
+				// use the ChMask of the first LinkADRReqPayload, besides
+				// turning off all 125 kHz this payload contains the ChMask
+				// for the last block of channels.
+				if c >= 64 {
+					payloads[0].ChMask[c%16] = true
+					println("Ch", c, true)
```
htdvisser: Don't `println`

core/networkserver/adr.go
```diff
@@ -168,36 +169,102 @@ func (n *networkServer) handleDownlinkADR(message *pb_broker.DownlinkMessage, de
...
+							ChMaskCntl: uint8(chMaskCntl),
+							NbRep: uint8(dev.ADR.NbTrans),
+						},
+					}
+
+					// set the channel mask for this block
+					for _, ec := range channels {
+						if ec >= chMaskCntl*16 && ec < (chMaskCntl+1)*16 {
+							pl.ChMask[ec%16] = true
+							println("Ch", c, true)
```
htdvisser: Don't `println`

core/band/band.go
```diff
@@ -90,6 +90,18 @@ func Get(region string) (frequencyPlan FrequencyPlan, err error) {
...
 		frequencyPlan.Band, err = lora.GetConfig(lora.US_902_928, false, lorawan.DwellTime400ms)
+		channelGroups := [](int){1} // Enable 903.9-905.3/200 kHz, 904.6 mHz channels
+		for channel := 0; channel < 72; channel++ {
+			for _, channelGroup := range channelGroups {
+				if (channel < channelGroup * 8 || channel >= (channelGroup + 1) * 8) && channel != channelGroup + 64 {
+					frequencyPlan.DisableUplinkChannel(channel)
+				}
+			}
+		}
+		// Is there a reason MaxDataRate shouldn't be 4?  I suppose if it was, the number of channels occupied would be
```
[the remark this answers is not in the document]
mgranberry: I dug into this a little more.  It looks like it only works as long as everything is ca. 3dB.  The noise floor increases are ~2.5 dB for each SF level and the EU power steps are 3dB.  The 3.5dB difference between SF7BW125 and SF8BW500 is great enough that I'm happy leaving it out, although it seems like it's probably worth experimenting with including the faster rates.

mgranberry:
@htdvisser This latest change adds support for the 2 dB TX power steps that are used in the US band after I identified an issue with some additional tests.  I believe I have now addressed all of the previous comments.

htdvisser:
@mgranberry: I couldn't push to your branch, so we'll continue with this in [feature/us-adr](https://github.com/TheThingsNetwork/ttn/tree/feature/us-adr)

PR #680, opened 2017-10-10
Add encrypted and fine timestamp to MQTT metadata

htdvisser:
Needed for localization


core/types/gateway_metadata.go
```diff
@@ -5,14 +5,16 @@ package types
 
 // GatewayMetadata contains metadata for each gateway that received a message
 type GatewayMetadata struct {
```
gomezjdaniel: What about the Frequency Offset and the other RSSI's?

htdvisser: According to our contact at Semtech those fields are not needed at this point.

PR #683, opened 2017-11-08
Mask sensitive information in error messages

johanstokking:
- Masking an app ID if the claims do not contain access to the app ID
- Masking an underlying error from the Account Server

### DIFF
--- a/core/handler/manager_server.go
+++ b/core/handler/manager_server.go
@@ -37,6 +37,9 @@ type handlerManager struct {
 }
 
 func checkAppRights(claims *claims.Claims, appID string, right types.Right) error {
+	if !claims.AppAccess(appID) {
+		return errors.NewErrPermissionDenied("No access to Application")
+	}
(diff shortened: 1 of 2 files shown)


core/handler/manager_server.go
```diff
@@ -55,7 +58,8 @@ func (h *handlerManager) validateTTNAuthAppContext(ctx context.Context, appID st
 		}
 		token, err := h.handler.Component.ExchangeAppKeyForToken(appID, key)
 		if err != nil {
-			return ctx, nil, err
+			h.handler.Ctx.WithError(err).Warn("Could not exchange key for token")
+			return ctx, nil, errors.NewErrPermissionDenied("Could not exchange key for token")
```
htdvisser: I think we should do this in the `ExchangeAppKeyForToken` func, as this error may either be caused by misconfiguration of the handler, or by an invalid access key. The first should be an InternalServerError, the second indeed a PermissionDenied

PR #691, opened 2017-12-22
Fatal error when a networkserver token can't be read

### LINKED ISSUE #690: Improper logging when the Broker NS token is invalid
When a broker's NS token is invalid, rather than the component failing immediately, every request fails with a `Warn`:

```
broker_1         | 2017-12-18T22:36:22.572402949Z   WARN Could not handle uplink                  ADR=false ADRAckReq=false Ack=false AppPayloadSize=6 Confirmed=false DevAddr=46000607 DownlinkOptions=2 Duplicates=1 FCnt=1698 FPending=false FPort=2 PayloadSize=19 error=NetworkServer did not return devices: Unable to verify JWT: illegal base64 data at input byte 81 source=Wrap: /go/src/github.com/TheThingsNetwork/ttn/utils/errors/errors.go:222
```

The broker should fail immediately at startup, such as by a pre-reading of the JWT, like it does with the discovery server token.

dotpy3:
Resolves #690.

### DIFF
--- a/cmd/broker.go
+++ b/cmd/broker.go
@@ -17,6 +17,7 @@ import (
 	ttnlog "github.com/TheThingsNetwork/go-utils/log"
 	"github.com/TheThingsNetwork/ttn/core/broker"
 	"github.com/TheThingsNetwork/ttn/core/component"
+	"github.com/dgrijalva/jwt-go"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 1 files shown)


htdvisser:
Merged in 33b404aff30fcfeba8a9c940d81139bb3e4532db

orybarfly:
Hi htdvisser,
I am trying to setup a private routing environment and I cannot proceed beyond the point when I have to register the prefix for the private networks. I am getting the message that the  "Auth server mynetwork-networkserver not registered".
I cannot figure out what have I done wrong during the previous lines of the setup, I've taken them one by one. The only point where I am in doubt is at the networkserver-token field of the ./broker/ttn.yml, I do not understand which token should I paste in the field. I have copied the same token as the one that I've entered in the auth-token field, I don't know if that is the reason that I cannot register the prefix.
Thank you in advance

johanstokking:
@orybarfly please consult Slack #support channel or the forum.

PR #699, opened 2018-03-03
Updated MQTT instructions as it uses Let's Encrypt since August 2017

avbentem:
Reference: https://www.thethingsnetwork.org/forum/t/secure-mqtt-apis-now-using-lets-encrypt-certificates/8774


johanstokking:
@avbentem thanks. Can you rebase?

avbentem:
Done.

PR #705, opened 2018-04-04
Removed ":1883" from mosquitto_sub examples

hraftery:
mosquitto_sub uses the `-p` flag to specify the port, or 1883 by default. Specifying it as part of the host name results in "Unable to connect (Lookup error.)".


johanstokking:
Thanks!

Issue #713, opened 2018-05-22
Add at least location as part of device events


johanstokking:
This is a **feature request** for **the backend**.

> What do you want to do?

Publish (at least) location as part of the device events.

This is especially relevant for the Collos integration, which updates the device registry and triggers a device update event. If we add the device's location as part of the payload, applications can subscribe to location updates through MQTT.

> What went wrong or what is missing?

Data is `nil`, see https://github.com/TheThingsNetwork/ttn/blob/develop/core/handler/manager_server.go#L228

toruuetani:
I think that Collos has three algorithm for resolving location.
So I think that update events need not only location but also algorithm.

johanstokking:
In V2 we have only one location per device in the registry.

Do note that the idea of Collos is that it resolves the best possible; in the order of WiFi, LoRa TDOA to LoRa RSSI. It is only for debugging purposes to have multiple endpoints enabled, but this is not supported in our device registry.

htdvisser:
Some notes for @Sypheos:
- There is no need to send the entire Device on every update. We can quite easily filter only the changed fields with `dev.ChangedFields()`
- We should **not** send sensitive fields such as keys, nonces, ...  
- We should probably send the jsonpb-marshaled (not std lib `encoding/json`!) for consistency with what grpc-gateway returns.

toruuetani:
OK, I understand.
We are interested only in TDOA.
I think that RSSI is TDOA's preliminary exercise.

johanstokking:
If you use the `Lora Wifi` recipe, you'll get the fallback automatically. Just don't pass WiFi; https://apiportal.preview.collos.org/docs/services/localization-lora-recipes/operations/59e9b2a367ba551b780cd259?

toruuetani:
OK, I understand.

toruuetani:
What topic does this event send to ?

johanstokking:
@toruuetani It will be part of the updated management event, see https://github.com/TheThingsNetwork/ttn/tree/develop/mqtt#management-events

toruuetani:
OK, thanks.

Sypheos:
@toruuetani Device data event now contain location PR #715 is merged

toruuetani:
@Sypheos Great !
I'm looking forward to evaluate this feature.

htdvisser:
I expect this to be deployed to production early next week 👍

toruuetani:
OK,  I will wait for update.
[...]

PR #715, opened 2018-05-29
Add updated data to device events.

Sypheos:
Add location and UpdateAt to device event data, if updated, when device a registry change.

Refer #713

### DIFF
--- a/core/handler/manager_server_test.go
+++ b/core/handler/manager_server_test.go
@@ -1,0 +1,43 @@
+package handler
+
+import (
+	"testing"
+	"time"
+
+	"github.com/TheThingsNetwork/ttn/core/handler/device"
+	. "github.com/smartystreets/assertions"
+)
+
+func TestEventUpdatedFields(t *testing.T) {
+	a := New(t)
+
+	dev := &device.Device{
+		AppID:     "test",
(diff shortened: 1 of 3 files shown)


core/handler/manager_server_test.go
```diff
@@ -0,0 +1,43 @@
+package handler
```
htdvisser: Missing header

core/types/event.go
```diff
@@ -81,3 +81,12 @@ type DownlinkEventData struct {
 	GatewayID string                   `json:"gateway_id,omitempty"`
 	Config    *DownlinkEventConfigInfo `json:"config,omitempty"`
 }
+
+// DeviceEventData contains the updated of a set/updated device
+type DeviceEventData struct {
+	Latitude   float32           `json:"latitude,omitempty"`
+	Longitude  float32           `json:"longitude,omitempty"`
+	Altitude   int32             `json:"altitude,omitempty"`
+	Attributes map[string]string `json:"attributes,omitempty"`
```
htdvisser: Not used?

Sypheos: yes forgot to remove it. It was first included but the `changeList` function doesn't detect change in the map.

PR #717, opened 2018-06-08
Add 1.0.2RevB Data Rates to AU915

htdvisser:
This PR overrides one of the vendored files in order to add the AU915 data rates that have been defined in 1.0.2RevB. 

I know this solution is a bit of a (dirty) workaround, but I think it's acceptable, considering that it saves a lot of time, allowing us to focus on v3 while still fixing the issue. Do you agree with this, @dotpy3?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ build-deps:
 
 deps: build-deps
 	govendor sync -v
+	git checkout vendor/github.com/brocaar/lorawan/band/band_au915_928.go
 
 dev-deps: deps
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -23,6 +23,7 @@ build-deps:
 
 deps: build-deps
 	govendor sync -v
+	git checkout vendor/github.com/brocaar/lorawan/band/band_au915_928.go
```
dotpy3: Won't this blow up if types in `github.com/brocaar/lorawan/band` go through breaking changes? Maybe we can specify a commit, or a tag?

htdvisser: That package is tagged by govendor, we simply override it with a checked-in version.

PR #742, opened 2018-10-18
Add support for RU_864_870 to ADR processing

belovictor:
This pull request adds support for RU_864_870 to adr.go

### DIFF
--- a/core/networkserver/adr.go
+++ b/core/networkserver/adr.go
@@ -332,6 +332,34 @@ func getAdrReqPayloads(dev *device.Device, frequencyPlan *band.FrequencyPlan, dr
 				}
 			}
 		}
+	case pb_lorawan.FrequencyPlan_RU_864_870.String():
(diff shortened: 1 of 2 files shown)


core/networkserver/adr.go
```diff
@@ -332,6 +332,34 @@ func getAdrReqPayloads(dev *device.Device, frequencyPlan *band.FrequencyPlan, dr
 				}
 			}
 		}
+	case pb_lorawan.FrequencyPlan_RU_864_870.String():
+		if dev.ADR.Failed > 0 && powerIdx > 5 {
```
htdvisser: This check was added to deal with EU868 devices that didn't support pwerIdx 6 and 7. Considering that `RU_864_870` already had powerIdx 0 - 7 support from the moment it was defined, I think we can leave this check out.

belovictor: OK!

PR #754, opened 2019-02-12
Make US and AU FSB configurable

### LINKED ISSUE #753: Make US915 FSB configurable
This is a **feature request** for **the backend**.

> What do you want to do?

Set a different FSB than 2 in US915.

> What steps did you take?

Set up a private network.

> What went wrong or what is missing?

Private network is fixed to FSB 2.

We probably need to make [this](https://github.com/TheThingsNetwork/ttn/blob/develop/core/band/band.go#L97) int configurable with default value `1` (zero-indexed).

htdvisser:
This PR resolves #753 in a similar way as #692 did for the RX2 Data Rate in the EU band. We introduce new flags `us-fsb` and `au-fsb` which default to 1 (the old hardcoded value), but can be changed by the network operator if required.

### DIFF
--- a/cmd/docs/README.md
+++ b/cmd/docs/README.md
@@ -6,6 +6,7 @@ The Things Network's backend servers.
 
 ```
       --allow-insecure                  Allow insecure fallback if TLS unavailable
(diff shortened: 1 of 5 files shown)


htdvisser:
cc: @Sypheos

PR #768, opened 2019-10-14
Do not fall back to SF12 when activated using OTAA

### LINKED ISSUE #767: RX2 fallback leads to endless downlinks
#### Summary
If a MAC command sent in RX2/**SF9** was not answered by an end node, the NS is using RX2/**SF12** as a fallback. In case that the end node IS listening on RX2/**SF9** and just missed the MAC command (due to temporarely unavailability or bad reachability) it will never receive the fallback MACs on RX2/**SF12**. In that case, again no answer is sent by the end node and the NS starts repeating the requests (LinkADRReq and RXParamSetupReq). That produces a huge amount of downlinks that will never be received by the end node that is still listening on RX2/**SF9**.

#### Steps to Reproduce
1. Start end node, join by OTAA
2. Wait for MAC command using RX2/**SF9**
3. Skip answer or shade end node in RX2 window
4. Watch MAC command using RX2/**SF12**
5. See endless repeated downlinks in RX2/**SF12**

#### What is already there? What do you see now?
grafik

#### What is missing? What do you want to see?
Not the same downlink again and again.

#### Environment
Not relevant.

#### How do you propose to implement this?
The RX2 fallback should not be used more than 2 times (for example). After 2 missed answeres, the RX2 should be changed back to **SF9** again. After 3 cycles of RX2/**SF9**-RX2/**SF12** changes some notification should be generated ("end node not answering any MAC") and the downlinks should be stopped. A restart of the procedure after an amount of time may be possible.

#### Can you do this yourself and submit a Pull Request?
No.

jpmeijers:
A modified version of @sigmaroot's fix (https://github.com/TheThingsNetwork/ttn/commit/edbe65d5a9a5e04c555166a0fa0b5fb4a712c5b7) was implemented and tested on a private local instance.

MQTT subscribe output of three different tests attached:
* Original version that incorrectly falls to SF12
* JP's idea of switching between SF12 and SF9 - not working
* Sigmaroot's fix with a small modification to get it to compile - tested to use SF9 to do downlinks. ie working fix.

[TTN private downlink with modified sigmaroot fix.txt](https://github.com/TheThingsNetwork/ttn/files/3726050/TTN.private.downlink.with.modified.sigmaroot.fix.txt)
[TTN private downlink with jp fix not working.txt](https://github.com/TheThingsNetwork/ttn/files/3726051/TTN.private.downlink.with.jp.fix.not.working.txt)
[TTN private downlink on sf12.txt](https://github.com/TheThingsNetwork/ttn/files/3726052/TTN.private.downlink.on.sf12.txt)

Fixes #767

### DIFF
--- a/core/networkserver/activation.go
+++ b/core/networkserver/activation.go
@@ -134,6 +134,7 @@ func (n *networkServer) HandleActivate(activation *pb_handler.DeviceActivationRe
 
 	dev.LastSeen = time.Now()
 	dev.UpdatedAt = time.Now()
+	dev.ActivatedAt = time.Now()
 	dev.DevAddr = *lorawan.DevAddr
 	dev.NwkSKey = *lorawan.NwkSKey
 	dev.FCntUp = 0
--- a/core/networkserver/device/device.go
+++ b/core/networkserver/device/device.go
@@ -36,9 +36,9 @@ type Device struct {
 	Options  Options       `redis:"options"`
 	ADR      ADRSettings   `redis:"adr,include"`
 
-	CreatedAt time.Time    `redis:"created_at"`
-	UpdatedAt time.Time    `redis:"updated_at"`
-	ActivatedAt time.Time  `redis:"activated_at"` // Indicates whether the device was activated via OTAA method
+	CreatedAt   time.Time `redis:"created_at"`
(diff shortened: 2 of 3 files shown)


htdvisser:
Thanks!

PR #769, opened 2019-12-31
Update README.md

DurgaPrasadWebDev:


htdvisser (changes requested):
Thanks for the improvement, @DurgaPrasadWebDev!

These docs are generated, so this line would actually need to be updated in

https://github.com/TheThingsNetwork/ttn/blob/develop/ttnctl/cmd/gateways_collaborators_add.go#L13

after which the docs can be regenerated with `make docs`.

johanstokking:
@DurgaPrasadWebDev will you update this PR? Thanks in advance!

htdvisser:
With The Things Network V2 shut down, we're going to archive this repository. This pull request will be locked.

PR #774, opened 2020-04-01
Add gRPC keep alive

adriansmares:
#### Changes

- Added the keep alive gRPC client call option with a period of 5 minutes. We use this value in order to conform to the default [enforcement policy](https://godoc.org/google.golang.org/grpc/keepalive#EnforcementPolicy), that we may encounter on external servers.
- Added the keep alive gRPC server option with a period of 30 seconds.
- Lowered the enforcement policy of the gRPC server to 1 ping per 30 seconds. Pings can occur in the absence of an active stream.
- Removed the `grpc.WithBlock` client call option.

#### Notes for Reviewers

The periods and timeouts are smaller than what I've proposed in v3. I don't have a strong opinion/argument for one value over the other one.

Should we keep the `grpc.WithBlock` ?  Does `restartstream` depend on it ?

### DIFF
--- a/api/pool/pool.go
+++ b/api/pool/pool.go
@@ -20,6 +20,7 @@ import (
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
 )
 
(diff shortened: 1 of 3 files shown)


htdvisser:
I'm fine with the values, don't think it matters much.

I don't think `restartstream` relies on `grpc.WithBlock`.

PR #776, opened 2020-04-29
Export v2 devices in v3 format

### LINKED ISSUE #764: Export devices as V3 JSON
This is a **feature request** for **ttnctl**.

> What do you want to do?

I want to migrate devices from v2 to v3.

> What steps did you take?

- In v2: `ttnctl devices list`, `ttnctl devices info [id]`
- In v3: `ttn-lw-cli end-devices create ...`

> What went wrong or what is missing?

It's too much work. It would be a lot easier if there was a `ttnctl devices export [id]` that would print V3 JSON so that I can pipe it directly into `ttn-lw-cli end-devices create`.

> Can you fix this yourself and submit a pull request?

Yes

neoaggelos:
## Summary

Closes #764 . This is the v2 part of migrating end devices to the v3.

## Changes

- Adds `ttnctl devices export <device-id>` command
- Adds `ttnctl devices export-all` command

## Example

```
$ ttnctl devices export-all --frequency-plan-id EU_863_870 | ttn-lw-cli dev create --application-id my-v3-app
```

It is required to pass the frequency plan ID (v2 devices don't have one). We could probably pick a default one from the handler/router, but I think it's better to let the user decide.

## Review

@htdvisser please review CLI and usage. Also, running `ttnctl` locally updates some packages in `go.mod`. Not sure how to proceed with these.
@rvolosatovs please review the MAC settings

### DIFF
--- a/ttnctl/cmd/devices_export.go
+++ b/ttnctl/cmd/devices_export.go
@@ -100,11 +100,10 @@ var devicesExportCmd = &cobra.Command{
 			ctx.WithError(err).Fatal("Could not get existing device.")
 		}
 
-		v3dev := exportV3Device(dev, cmd.Flags())
-
 		format, _ := cmd.Flags().GetString("format")
 		switch format {
 		case "v3":
+			v3dev := exportV3Device(dev, cmd.Flags())
 			if err := json.NewEncoder(os.Stdout).Encode(v3dev); err != nil {
 				conn.Close()
 				ctx.WithError(err).Fatal("Could not export device in v3 format")
@@ -133,11 +132,10 @@ var devicesExportAllCmd = &cobra.Command{
 		}
 
 		for _, dev := range devs {
-			v3dev := exportV3Device(dev, cmd.Flags())
-
 			format, _ := cmd.Flags().GetString("format")
 			switch format {
 			case "v3":
+				v3dev := exportV3Device(dev, cmd.Flags())
 				if err := json.NewEncoder(os.Stdout).Encode(v3dev); err != nil {
 					conn.Close()
 					ctx.WithError(err).Fatal("Could not export device in v3 format")
--- a/ttnctl/cmd/devices_export.go
+++ b/ttnctl/cmd/devices_export.go
@@ -1,0 +1,163 @@
+// Copyright © 2020 The Things Network
+// Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+
+package cmd
+
+import (
+	"encoding/json"
+	"os"
+	"strings"
+
+	"github.com/TheThingsNetwork/api"
+	"github.com/TheThingsNetwork/api/handler"
+	"github.com/TheThingsNetwork/go-utils/log"
+	"github.com/TheThingsNetwork/ttn/ttnctl/util"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+func exportV3Device(dev *handler.Device, flags *pflag.FlagSet) *v3Device {
+	v3Dev := &v3Device{}
+	v3Dev.IDs.ApplicationIDs.ApplicationID = dev.AppID
+	v3Dev.IDs.DeviceID = dev.DevID
+
+	if dev.Latitude != 0 || dev.Longitude != 0 {
+		v3Dev.Locations = map[string]v3DeviceLocation{
+			"user": {
+				Latitude:  float64(dev.Latitude),
+				Longitude: float64(dev.Longitude),
+				Altitude:  dev.Altitude,
+				Source:    "SOURCE_REGISTRY",
+			},
+		}
+	}
+
+	v3Dev.Name = dev.DevID
+	v3Dev.Description = dev.Description
+	v3Dev.Attributes = dev.Attributes
+
+	lorawanDevice := dev.GetLoRaWANDevice()
+
+	v3Dev.IDs.JoinEUI = lorawanDevice.GetAppEUI().String()
+	v3Dev.IDs.DevEUI = lorawanDevice.GetDevEUI().String()
+	if lorawanDevice.AppKey.String() != "" {
+		v3Dev.RootKeys = &v3DeviceRootKeys{
+			AppKey: v3DeviceKey{Key: lorawanDevice.AppKey.String()},
+		}
+	}
+	if devAddr := lorawanDevice.DevAddr; !devAddr.IsEmpty() && lorawanDevice.NwkSKey.String() != "" {
+		v3Dev.DevAddr = devAddr.String()
+		v3Dev.Session = &v3DeviceSession{}
+		v3Dev.Session.DevAddr = lorawanDevice.DevAddr.String()
+		if lorawanDevice.AppSKey.String() != "" {
+			v3Dev.Session.Keys.AppSKey = v3DeviceKey{Key: lorawanDevice.AppSKey.String()}
+		}
+		v3Dev.Session.Keys.FNwkSIntKey = v3DeviceKey{Key: lorawanDevice.NwkSKey.String()}
+		v3Dev.Session.LastFCntUp = lorawanDevice.FCntUp
+		v3Dev.Session.LastNFCntDown = lorawanDevice.FCntDown
+	}
+
+	v3Dev.MACSettings.Supports32BitFCnt = lorawanDevice.Uses32BitFCnt
+	v3Dev.MACSettings.ResetsFCnt = lorawanDevice.DisableFCntCheck
+
+	v3Dev.MACVersion = "MAC_V1_0_2"
+	v3Dev.PHYVersion = "PHY_V1_0_2_REV_B"
+
+	v3Dev.SupportsJoin = lorawanDevice.AppKey.String() != ""
+	v3Dev.MACSettings.Rx1Delay = "RX_DELAY_1"
+
+	frequencyPlanFlag, _ := flags.GetString("frequency-plan-id")
+	frequencyPlan, err := getOption(frequencyPlans, frequencyPlanFlag)
+	if err != nil {
+		ctx.WithError(err).WithFields(log.Fields{
+			"frequency_plan_id":            frequencyPlanFlag,
+			"available_frequency_plan_ids": frequencyPlans,
+		}).Fatal("Invalid --frequency-plan-id argument")
+	}
+	v3Dev.FrequencyPlanID = frequencyPlan
+
+	return v3Dev
+}
+
+var devicesExportCmd = &cobra.Command{
+	Use:     "export [Device ID]",
+	Short:   "Export a device",
+	Long:    `ttnctl devices export exports a device to an external format.`,
+	Example: `$ ttnctl devices export test | ttn-lw-cli end-devices create --application-id app-id`,
+	Run: func(cmd *cobra.Command, args []string) {
+		assertArgsLength(cmd, args, 1, 1)
+
+		devID := strings.ToLower(args[0])
+		if err := api.NotEmptyAndValidID(devID, "Device ID"); err != nil {
+			ctx.Fatal(err.Error())
+		}
+		appID := util.GetAppID(ctx)
+
+		conn, manager := util.GetHandlerManager(ctx, appID)
+		dev, err := manager.GetDevice(appID, devID)
+		if err != nil {
+			conn.Close()
+			ctx.WithError(err).Fatal("Could not get existing device.")
+		}
+
+		v3dev := exportV3Device(dev, cmd.Flags())
+
+		format, _ := cmd.Flags().GetString("format")
+		switch format {
+		case "v3":
+			if err := json.NewEncoder(os.Stdout).Encode(v3dev); err != nil {
+				conn.Close()
+				ctx.WithError(err).Fatal("Could not export device in v3 format")
+			}
+		}
+		conn.Close()
+	},
+}
+
+var devicesExportAllCmd = &cobra.Command{
+	Use:     "export-all",
+	Short:   "Export all devices",
+	Long:    `ttnctl devices export-all exports all devices to an external format.`,
+	Example: `$ ttnctl devices export-all | ttn-lw-cli end-devices create --application-id app-id`,
+	Run: func(cmd *cobra.Command, args []string) {
+		assertArgsLength(cmd, args, 0, 0)
+
+		appID := util.GetAppID(ctx)
+
+		conn, manager := util.GetHandlerManager(ctx, appID)
+
+		devs, err := manager.GetDevicesForApplication(appID, 0, 0)
+		if err != nil {
+			conn.Close()
+			ctx.WithError(err).Fatal("Could not get devices.")
+		}
+
+		for _, dev := range devs {
+			v3dev := exportV3Device(dev, cmd.Flags())
+
+			format, _ := cmd.Flags().GetString("format")
+			switch format {
+			case "v3":
+				if err := json.NewEncoder(os.Stdout).Encode(v3dev); err != nil {
+					conn.Close()
+					ctx.WithError(err).Fatal("Could not export device in v3 format")
+				}
+			}
+		}
+		conn.Close()
+	},
+}
+
+func v3ExportFlagSet() *pflag.FlagSet {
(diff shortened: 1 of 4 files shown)


ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		ctx.WithError(err).WithFields(log.Fields{
+			"frequency_plan_id":            frequencyPlanFlag,
+			"available_frequency_plan_ids": frequencyPlans,
+		}).Fatal("Invalid --frequency-plan-id argument")
+	}
+	v3Dev.FrequencyPlanID = frequencyPlan
+
+	v3Dev.SupportsClassB, _ = flags.GetBool("class-b")
+	v3Dev.SupportsClassC, _ = flags.GetBool("class-c")
```
htdvisser, line 89: Can remove these, because they are always false. V2 never had support for class B and C.

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+	},
+}
+
+func v3ExportFlagSet() *pflag.FlagSet {
+	set := &pflag.FlagSet{}
+	set.String("format", "v3", "Formatting: v3")
+	set.String("frequency-plan-id", "", "Specify Frequency Plan ID")
+	set.Bool("class-b", false, "Enable Class B")
+	set.Bool("class-c", false, "Enable Class C")
```
htdvisser, line 162: Remove

ttnctl/cmd/devices_export.go
neoaggelos: Done.

ttnctl/cmd/devices_export.go
neoaggelos: Done.

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,166 @@
...
+	lorawanDevice := dev.GetLoRaWANDevice()
+
+	v3Dev.IDs.JoinEUI = lorawanDevice.GetAppEUI().String()
+	v3Dev.IDs.DevEUI = lorawanDevice.GetDevEUI().String()
+	if lorawanDevice.AppKey.String() != "" {
+		v3Dev.RootKeys = &v3DeviceRootKeys{
+			AppKey: v3DeviceKey{Key: lorawanDevice.AppKey.String()},
+		}
+	}
+	if devAddr := lorawanDevice.DevAddr; !devAddr.IsEmpty() {
```
rvolosatovs, line 48:
```suggestion
	if devAddr := lorawanDevice.DevAddr; !devAddr.IsEmpty() && lorawanDevice.NwkSKey.String() != "" {
```

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,166 @@
...
+		v3Dev.DevAddr = devAddr.String()
+		v3Dev.Session = &v3DeviceSession{}
+		v3Dev.Session.DevAddr = lorawanDevice.DevAddr.String()
+		if lorawanDevice.AppSKey.String() != "" {
+			v3Dev.Session.Keys.AppSKey = v3DeviceKey{Key: lorawanDevice.AppSKey.String()}
+		}
+		if lorawanDevice.NwkSKey.String() != "" {
+			v3Dev.Session.Keys.FNwkSIntKey = v3DeviceKey{Key: lorawanDevice.NwkSKey.String()}
+		}
```
rvolosatovs, line 57:
```suggestion
		v3Dev.Session.Keys.FNwkSIntKey = v3DeviceKey{Key: lorawanDevice.NwkSKey.String()}
```

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,166 @@
...
+
+	if lorawanDevice.AppKey.String() == "" {
+		v3Dev.SupportsJoin = false
+		v3Dev.MACSettings.Rx1Delay = "RX_DELAY_1"
+	} else {
+		v3Dev.SupportsJoin = true
+		if !lorawanDevice.AppSKey.IsEmpty() && !lorawanDevice.NwkSKey.IsEmpty() && !lorawanDevice.DevAddr.IsEmpty() {
+			v3Dev.MACState.CurrentParameters.Rx1Delay = "RX_DELAY_1"
+		}
+	}
```
rvolosatovs, line 76:
```suggestion
	v3Dev.SupportsJoin = lorawanDevice.AppKey.String() != ""
	v3Dev.MACSettings.Rx1Delay = "RX_DELAY_1"
```
1. Unfortunately Network Server does not support creating devices with MAC state at this point (it gets overwritten) - it's easy to fix and I will open an issue about it
2. Using RX Delay of 1 will give users more familiar experience, RX delay of 5 might in fact break some use cases, so IMO this is the correct approach. @johanstokking what do you think?

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,166 @@
...
+
+	frequencyPlanFlag, _ := flags.GetString("frequency-plan-id")
+	frequencyPlan, err := getOption(frequencyPlans, frequencyPlanFlag, "")
+	if err != nil {
+		ctx.WithError(err).WithFields(log.Fields{
+			"frequency_plan_id":            frequencyPlanFlag,
+			"available_frequency_plan_ids": frequencyPlans,
+		}).Fatal("Invalid --frequency-plan-id argument")
+	}
+	v3Dev.FrequencyPlanID = frequencyPlan
```
rvolosatovs, line 86: What would be nice to have here is also validating this according to https://github.com/TheThingsNetwork/lorawan-frequency-plans
@htdvisser are these:
1. validated
2. using the same IDs

?

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,166 @@
...
+			ctx.WithError(err).Fatal("Could not get existing device.")
+		}
+
+		v3dev := exportV3Device(dev, cmd.Flags())
+
+		format, _ := cmd.Flags().GetString("format")
+		switch format {
+		case "v3":
+			b, _ := json.Marshal(v3dev)
+			fmt.Println(string(b))
```
rvolosatovs, line 118:
```suggestion
			if err := json.NewEncoder(os.Stdout).Encode(v3dev); err != nil {
				ctx.WithError(err).Fatal("Could not export device in v3 format to stdout.")
			}
```

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,166 @@
...
+			ctx.WithError(err).Fatal("Could not get devices.")
+		}
+
+		for _, dev := range devs {
+			v3dev := exportV3Device(dev, cmd.Flags())
+
+			format, _ := cmd.Flags().GetString("format")
+			switch format {
+			case "v3":
+				b, _ := json.Marshal(v3dev)
```
rvolosatovs, line 147: IDEM

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,166 @@
...
+		assertArgsLength(cmd, args, 1, 1)
+
+		devID := strings.ToLower(args[0])
+		if err := api.NotEmptyAndValidID(devID, "Device ID"); err != nil {
+			ctx.Fatal(err.Error())
+		}
+		appID := util.GetAppID(ctx)
+
+		conn, manager := util.GetHandlerManager(ctx, appID)
+		defer conn.Close()
```
rvolosatovs, line 106: This is ignored on `ctx.Fatal` calls
Either extract this functionality into a separate function not using `Fatal`, but simply returning an error instead or add this call before each `Fatal` call and possible return

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,166 @@
...
+	Short:   "Export all devices",
+	Long:    `ttnctl devices export-all exports all devices to an external format.`,
+	Example: `$ ttnctl devices export-all | ttn-lw-cli end-devices create --application-id app-id`,
+	Run: func(cmd *cobra.Command, args []string) {
+		assertArgsLength(cmd, args, 0, 0)
+
+		appID := util.GetAppID(ctx)
+
+		conn, manager := util.GetHandlerManager(ctx, appID)
+		defer conn.Close()
```
rvolosatovs, line 134: IDEM

ttnctl/cmd/root.go
```diff
@@ -38,9 +38,16 @@ var RootCmd = &cobra.Command{
 			logLevel = log.DebugLevel
 		}
 
+		logWriter := os.Stdout
+		if cmd == devicesExportCmd {
+			logWriter = os.Stderr
+		} else {
+			fmt.Printf("%#v\n", cmd)
```
rvolosatovs, line 45: What does this do exactly?

ttnctl/cmd/v3types.go
```diff
@@ -0,0 +1,104 @@
+// Copyright © 2020 The Things Network
+// Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+
+package cmd
+
+import "fmt"
+
+type v3ApplicationIDs struct {
```
rvolosatovs, line 8: If we had https://github.com/TheThingsNetwork/lorawan-stack/pull/2453 we could just import the `ttnpb` types (cc @johanstokking )

ttnctl/cmd/devices_export.go
rvolosatovs: Without `FNwkSIntKey` (aka `NwkSKey`) Network Server cannot use the session

ttnctl/cmd/devices_export.go
johanstokking: Migrated device sessions have to respect the existing RX delay of 1 second, but I'm open to increasing it in the desired parameters.

Maybe I misunderstood your question. In that case, please clarify

ttnctl/cmd/devices_export.go
johanstokking: These are not necessarily the same identifiers I believe

ttnctl/cmd/v3types.go
johanstokking: I don't think we want to import the V3 repo in V2

ttnctl/cmd/root.go
johanstokking: Seems like a debugging leftover

ttnctl/cmd/devices_export.go
rvolosatovs: My suggestion is to explicitly use RX delay of 1 in MAC settings(i.e. always) for all migrated devices, see suggested diff.

ttnctl/cmd/devices_export.go
neoaggelos: Calling conn.Close() on all possible exit points.

ttnctl/cmd/devices_export.go
neoaggelos: I took the IDs from the github repo. I think it's overkill to dynamically retrieve them every time (they barely change anyway)

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,172 @@
...
+	v3Dev.MACSettings.Rx1Delay = "RX_DELAY_1"
+
+	if lorawanDevice.AppKey.String() == "" {
+		v3Dev.SupportsJoin = false
+	} else {
+		v3Dev.SupportsJoin = true
+		if !lorawanDevice.AppSKey.IsEmpty() && !lorawanDevice.NwkSKey.IsEmpty() && !lorawanDevice.DevAddr.IsEmpty() {
+			v3Dev.MACState.CurrentParameters.Rx1Delay = "RX_DELAY_1"
+		}
+	}
```
rvolosatovs, line 76:
```suggestion
	if lorawanDevice.AppKey.String() == "" {
		v3Dev.SupportsJoin = false
	} else {
		v3Dev.SupportsJoin = true
		if !lorawanDevice.AppSKey.IsEmpty() && !lorawanDevice.NwkSKey.IsEmpty() && !lorawanDevice.DevAddr.IsEmpty() {
			v3Dev.MACState.CurrentParameters.Rx1Delay = "RX_DELAY_1"
		}
	}
```

This whole section can be removed, as suggested before already

ttnctl/cmd/v3types.go
```diff
@@ -0,0 +1,104 @@
...
+	SupportsJoin    bool                        `json:"supports_join"`
+	RootKeys        *v3DeviceRootKeys           `json:"root_keys,omitempty"`
+	NetID           []byte                      `json:"net_id,omitempty"`
+	MACSettings     v3DeviceMACSettings         `json:"mac_settings"`
+	MACState        v3DeviceMACState            `json:"mac_state"`
+	Session         *v3DeviceSession            `json:"session,omitempty"`
+}
+
+var (
+	frequencyPlans = []string{
```
rvolosatovs, line 77: nit; please sort these alphabetically, it's much easier to compare them to contents of github repository that way.
*(in vim it's just `:sort` command, your editor of choice should support that as well, if it does not - find a new editor)*

ttnctl/cmd/v3types.go
```diff
@@ -0,0 +1,104 @@
...
+		"EU_863_870",
+		"EU_863_870_TTN",
+		"US_902_928_FSB_1",
+		"US_902_928_FSB_2",
+		"AU_915_928_FSB_1",
+		"AU_915_928_FSB_2",
+		"AU_915_928_FSB_6",
+		"CN_470_510_FSB_11",
+		"AS_920_923",
+		"AS_920_923_LBT",
```
rvolosatovs, line 87: I do not see this frequency plan at https://github.com/TheThingsNetwork/lorawan-frequency-plans/tree/f113215f81295147cae0d1bd74c2a665529630ac

ttnctl/cmd/v3types.go
```diff
@@ -0,0 +1,104 @@
...
+		"AS_923_925",
+		"AS_923_925_LBT",
```
rvolosatovs, line 89: IDEM

ttnctl/cmd/devices_export.go
neoaggelos: Fixed, missed on the previous commit

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,172 @@
...
+
+	frequencyPlanFlag, _ := flags.GetString("frequency-plan-id")
+	frequencyPlan, err := getOption(frequencyPlans, frequencyPlanFlag, "")
```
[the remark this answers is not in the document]
neoaggelos: It does error, and also lists the available frequency plan ids. Do I understand correctly?

ttnctl/cmd/v3types.go
```diff
@@ -0,0 +1,99 @@
...
+		"EU_863_870",
+		"IN_865_867",
+		"KR_920_923_TTN",
+		"RU_864_870_TTN",
+		"US_902_928_FSB_1",
+		"US_902_928_FSB_2",
+	}
+)
+
+func getOption(options []string, option, fallback string) (string, error) {
```
rvolosatovs, line 92:
```suggestion
func getOption(options []string, option string) (string, error) {
```

Normally all return values besides the `error` are garbage when `error` is not nil. Having a fallback here is simply confusing, since it's only returned on errors (when a `""` should be returned)
See https://github.com/TheThingsNetwork/ttn/pull/776/files#r418098866

neoaggelos: Ah, yes, another leftover before settling with this implementation, removed

ttnctl/cmd/devices_export.go
```diff
@@ -0,0 +1,163 @@
...
+			conn.Close()
+			ctx.WithError(err).Fatal("Could not get existing device.")
+		}
+
+		v3dev := exportV3Device(dev, cmd.Flags())
+
+		format, _ := cmd.Flags().GetString("format")
+		switch format {
+		case "v3":
+			if err := json.NewEncoder(os.Stdout).Encode(v3dev); err != nil {
```
rvolosatovs, line 108: nit; since you have this switch, it actually makes more sense to export to v3 format inside the `v3` case, not outside the switch, same below.

PR #777, opened 2020-05-06
cli: Add ttnctl devices convert-to-abp command

neoaggelos:
## Summary

Needed for https://github.com/TheThingsNetwork/lorawan-stack/pull/2467

## Changes

Add a new `ttnctl devices disable` command that resets AppKey of OTAA devices.

Also adds a `ttnctl devices disable-all` variant.

The user can pass a `--save-to-attribute <attribute-name>` to save the original AppKey as an attribute on the device (if they want to). The original AppKey is printed on the command output in any case.

## Notes

- Device AppKey is changed to 00...001. It cannot be zero or empty, because the device will then be treated as an ABP device. Do we want that?

Some more ideas I played around with but left out for now:

- Should we do something with ABP devices? E.g. reset AppSKey or something?

### DIFF
--- a/ttnctl/cmd/devices_convert_to_abp.go
+++ b/ttnctl/cmd/devices_convert_to_abp.go
@@ -1,0 +1,125 @@
+// Copyright © 2017 The Things Network
+// Use of this source code is governed by the MIT license that can be found in the LICENSE file.
+
+package cmd
+
+import (
+	"strings"
+
+	"github.com/TheThingsNetwork/api"
+	"github.com/TheThingsNetwork/api/handler"
+	"github.com/TheThingsNetwork/api/handler/handlerclient"
+	ttnlog "github.com/TheThingsNetwork/go-utils/log"
+	"github.com/TheThingsNetwork/ttn/core/types"
+	"github.com/TheThingsNetwork/ttn/ttnctl/util"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+func convertToABP(manager *handlerclient.ManagerClient, dev *handler.Device, flags *pflag.FlagSet) {
+	// Do all updates
+	if dev.GetLoRaWANDevice().AppKey.String() != "" {
+		old := dev.GetLoRaWANDevice().AppKey.String()
+		dev.GetLoRaWANDevice().AppKey = &types.AppKey{}
+
+		attr, _ := flags.GetString("save-to-attribute")
+		if attr != "" {
+			if dev.Attributes == nil {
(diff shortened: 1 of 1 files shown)


ttnctl/cmd/devices_disable.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/TheThingsNetwork/ttn/ttnctl/util"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+func disableDevice(manager *handlerclient.ManagerClient, dev *handler.Device, flags *pflag.FlagSet) {
+	// Do all updates
+	if dev.GetLoRaWANDevice().AppKey.String() != "" {
+		old := dev.GetLoRaWANDevice().AppKey.String()
+		dev.GetLoRaWANDevice().AppKey = &types.AppKey{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01}
```
htdvisser, line 23: Setting a key to something constant and known is a bad idea.

neoaggelos: Now setting empty AppKey

neoaggelos:
Renamed commands to `convert-to-abp` and `convert-all-to-abp`.

I just want the commands to make sense in the context of the migration guide.

ttnctl/cmd/devices_convert_to_abp.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/TheThingsNetwork/ttn/ttnctl/util"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+func convertToABP(manager *handlerclient.ManagerClient, dev *handler.Device, flags *pflag.FlagSet) {
+	// Do all updates
+	if dev.GetLoRaWANDevice().AppKey.String() != "" {
+		old := dev.GetLoRaWANDevice().AppKey.String()
+		dev.GetLoRaWANDevice().AppKey = &types.AppKey{}
```
johanstokking, line 23: The AppKey is still set, but empty. This allows anyone to activate the device with this known, empty AppKey.

Right? Or do I misunderstand how this field is interpreted?

Why not setting this field to `nil`?

neoaggelos: That would make sense, but it does not work like that.

Setting `AppKey` to nil here has no effect on the device. Setting it to empty, actually clears it and makes the device appear as ABP on the console.

PR #792, opened 2021-08-25
Allow clearing session keys

KrishnaIyer:
#### Summary

Allow clearing session keys, this is needed by the migration tool.

#### Changes

- Check the changed list for `AppSKey` and `NwkSKey` and allow update.

#### Testing

🤷   let's see if this breaks anything once deployed.

### DIFF
--- a/core/handler/manager_server.go
+++ b/core/handler/manager_server.go
@@ -225,7 +225,7 @@ func (h *handlerManager) SetDevice(ctx context.Context, in *pb_handler.Device) (
 		changedFields := dev.ChangedFields()
 		for _, changedField := range changedFields {
 			switch changedField {
-			case "AppKey", "UsedDevNonces", "UsedAppNonces": // Allow changing the AppKey.
+			case "AppKey", "UsedDevNonces", "UsedAppNonces", "NwkSKey", "AppSKey": // Allow changing the AppKey for OTAA and the session keys for ABP.
 			default:
 				md, _ := metadata.FromIncomingContext(ctx)
 				h.handler.Ctx.WithFields(ttnlog.Fields{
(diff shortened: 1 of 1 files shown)


adriansmares (approved):

KrishnaIyer:
This is deployed to all TTN foundation clusters. Waiting for feedback.

KrishnaIyer:
So this also needs `DevAddr`; https://github.com/TheThingsNetwork/lorawan-stack-migrate/blob/67b4b43c12c4ff756799cc89b3f1eb2ba6504100/pkg/source/ttnv2/source.go#L176

KrishnaIyer:
This is deployed and tested, currently running [my dev image](https://hub.docker.com/layers/krishnaiyer/ttn/03a6c7ca/images/sha256-b9a35de3105237d1523b633f27aeaab1ba77a9f4776e3af3155a71cc7a2813a0).

adriansmares (approved):

core/handler/manager_server.go
```diff
@@ -225,7 +225,8 @@ func (h *handlerManager) SetDevice(ctx context.Context, in *pb_handler.Device) (
 		changedFields := dev.ChangedFields()
 		for _, changedField := range changedFields {
 			switch changedField {
-			case "AppKey", "UsedDevNonces", "UsedAppNonces": // Allow changing the AppKey.
+			case "AppKey", "UsedDevNonces", "DevAddr", "UsedAppNonces", "NwkSKey", "AppSKey": // Allow changing the AppKey for OTAA and the session data for ABP.
+			case "FCntUp":
```
htdvisser: Is this a leftover from "just trying something" or is this intended to be a separate case?

adriansmares: This is intended:
- In v2 there are no field masks, so you need to patch the full device.
- If the device sends an uplink between a `Get` and ` Set`, the Handler detects that you're trying to update the frame counter and fails.

Do you see a better fix ?

htdvisser: Okay, fine with me, but let's add a quick comment. Can you do that?

htdvisser (approved):

